PR #9, opened 2023-10-17
CDR from `pm_icecon`

trey-stafford:
Brings in the CLI and other logic specific to the CDR from `pm_icecon`. See the associated PR for that repo here: https://github.com/nsidc/pm_icecon/pull/27

### DIFF
--- a/seaice_ecdr/cdr_alg.py
+++ b/seaice_ecdr/cdr_alg.py
@@ -34,7 +34,6 @@ import pm_icecon.nt.params.amsr2 as nt_amsr2_params
 from pm_icecon._types import Hemisphere
 from pm_icecon.cli.util import datetime_to_date
 from pm_icecon.config.models.bt import BootstrapParams
-from pm_icecon.constants import CDR_DATA_DIR, DEFAULT_FLAG_VALUES
 from pm_icecon.fill_polehole import fill_pole_hole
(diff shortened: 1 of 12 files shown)


trey-stafford:
Note that we expect tests to fail on this branch until  https://github.com/nsidc/pm_icecon/pull/27 is merged.

seaice_ecdr/cdr_alg.py
trey-stafford, line 1: This was `pm_icecon.cdr`. Just moved wholesale to this repo and renamed `cdr_alg`. This needs further cleanup but will come in another PR.

seaice_ecdr/compare/__init__.py
trey-stafford, line 1: The comparison/viz code is specific to the CDR's needs so it's moved over here for now.

seaice_ecdr/masks.py
```diff
@@ -0,0 +1,24 @@
+import numpy as np
+import numpy.typing as npt
+
+
+def psn_125_near_pole_hole_mask() -> npt.NDArray[np.bool_]:
```
trey-stafford, line 5: Extracted the code that creates the 'near pole mask' to this function so that the `fill_polehole` function could be more generic (takes a `near_pole_hole_mask` as a kwarg)

sc0tts (approved):
So much better to have the logic properly segregated!

PR #11, opened 2023-10-18
NRT LANCE AMSR2 Initial daily file creation

trey-stafford:
Support creating initial daily files from NRT LANCE AMSR2 data.

### DIFF
--- a/seaice_ecdr/constants.py
+++ b/seaice_ecdr/constants.py
@@ -1,12 +1,15 @@
 from pathlib import Path
 
+# This is the version string for the ECDR product.
+ECDR_PRODUCT_VERSION = "v5"
+
 # NSIDC infrastructure-specific paths:
 NSIDC_NFS_SHARE_DIR = Path("/share/apps/amsr2-cdr")
 
 # TODO: dev-specific directories for the outputs!
 
 # Outputs from the `seaice_ecdr` go to these locations.
-BASE_OUTPUT_DIR = NSIDC_NFS_SHARE_DIR / "ecdr_outputs"
+BASE_OUTPUT_DIR = NSIDC_NFS_SHARE_DIR / f"ecdr_{ECDR_PRODUCT_VERSION}_outputs"
 # Daily initial/intermiedate output for 'standard' (not NRT) ECDR processing
 # (e.g, using input data from AU_SI12)
 INITIAL_DAILY_OUTPUT_DIR = BASE_OUTPUT_DIR / "standard" / "initial_daily"
--- a/seaice_ecdr/constants.py
+++ b/seaice_ecdr/constants.py
@@ -3,7 +3,15 @@ from pathlib import Path
 # NSIDC infrastructure-specific paths:
 NSIDC_NFS_SHARE_DIR = Path("/share/apps/amsr2-cdr")
 
+# Outputs from the `seaice_ecdr` go to these locations.
+BASE_OUTPUT_DIR = NSIDC_NFS_SHARE_DIR / "ecdr_outputs"
+# Daily initial output from 'final' input data (e.g., AU_SI12)
+INITIAL_DAILY_OUTPUT_DIR = BASE_OUTPUT_DIR / "final" / "initial_daily"
+# Daily initial output from NRT LANCE AMSR2 data (AU_SI12_NRT_R04)
+NRT_INITIAL_DAILY_OUTPUT_DIR = BASE_OUTPUT_DIR / "nrt" / "initial_daily"
+
 # Location of LANCE AMSR2 NRT data files:
+# TODO: nest the subdir under an `ecdr_inputs` or similar?
 LANCE_NRT_DATA_DIR = NSIDC_NFS_SHARE_DIR / "lance_amsr2_nrt_data"
 
 CDR_DATA_DIR = NSIDC_NFS_SHARE_DIR / "cdr_data"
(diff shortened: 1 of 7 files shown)


seaice_ecdr/constants.py
```diff
@@ -3,7 +3,15 @@
 # NSIDC infrastructure-specific paths:
 NSIDC_NFS_SHARE_DIR = Path("/share/apps/amsr2-cdr")
 
+# Outputs from the `seaice_ecdr` go to these locations.
+BASE_OUTPUT_DIR = NSIDC_NFS_SHARE_DIR / "ecdr_outputs"
+# Daily initial output from 'final' input data (e.g., AU_SI12)
+INITIAL_DAILY_OUTPUT_DIR = BASE_OUTPUT_DIR / "final" / "initial_daily"
```
sc0tts, line 9: This is okay for now, but the use of "final" for "intermediate" files threw me off a bit.  It's because the "final" here refers to the input, not the ecdr files.

Maybe a TODO that labels inputs from AUSI12 "final" as ... AUSI12_B04 ?  Or AUSI12_pub[lished]?

This is TBD.  Not blocking this PR.

seaice_ecdr/constants.py
```diff
@@ -3,7 +3,15 @@
...
+# Daily initial output from NRT LANCE AMSR2 data (AU_SI12_NRT_R04)
+NRT_INITIAL_DAILY_OUTPUT_DIR = BASE_OUTPUT_DIR / "nrt" / "initial_daily"
+
 # Location of LANCE AMSR2 NRT data files:
+# TODO: nest the subdir under an `ecdr_inputs` or similar?
 LANCE_NRT_DATA_DIR = NSIDC_NFS_SHARE_DIR / "lance_amsr2_nrt_data"
 
 CDR_DATA_DIR = NSIDC_NFS_SHARE_DIR / "cdr_data"
```
sc0tts, line 17: Here, we might want to call out the version of CDR data:  so "cdrv5_data" so that it's not confused with other concurrent processes

seaice_ecdr/initial_daily_ecdr.py
```diff
@@ -176,34 +180,27 @@ def calculate_bt_nt_cdr_raw_conc(
     return bt_conc, nt_conc, cdr_conc
 
 
-def compute_initial_daily_ecdr_dataset(
+def _setup_ecdr_ds(
```
sc0tts, line 183: I'm influenced by my prior comment.  We've been referring to this as the "eCDR" but it will actually be delivered as "CDRv5" (I think).  At some point, we will probably want to refactor out the label "eCDR" for clarity of later developers who weren't involved in this transition.

seaice_ecdr/initial_daily_ecdr.py
```diff
@@ -217,15 +214,8 @@ def compute_initial_daily_ecdr_dataset(
...
-    # Get AU_SI TBs
-    xr_tbs = get_au_si_tbs(
-        date=date,
-        hemisphere=hemisphere,
-        resolution=resolution,
-    )
-
     # Move TBs to ecdr_ds
-    for tbname in ("h18", "v18", "v23", "h36", "v36"):
+    for tbname in EXPECTED_TB_NAMES:
```
sc0tts, line 218: Another TODO for later:  Eventually, we do want to refactor these TB names into "channel" names.  Because the exact frequency depends on the platform.

This is another of those fairly major refactorings that will clarify the scientific reasoning, but should not be undertaken until we have stable examples that can used to verify that these code changes do not introduce numerical differences in the outputs.

seaice_ecdr/initial_daily_ecdr.py
```diff
@@ -248,13 +238,42 @@ def compute_initial_daily_ecdr_dataset(
...
+
+
+def compute_initial_daily_ecdr_dataset(
+    *,
+    date: dt.date,
+    hemisphere: Hemisphere,
+    resolution: AU_SI_RESOLUTIONS,
+    xr_tbs: xr.Dataset,
+):
+    ecdr_ide_ds = _setup_ecdr_ds(
```
sc0tts, line 265: I only notice this occasionally, but we should probably have a docstring at the start of these large-scale functions.

sc0tts (approved):
I added a lot of "TODO" suggestions, but did not see anything that prevents me from recommending merging this PR.  Nice work!

seaice_ecdr/constants.py
trey-stafford: Yea, agreed! I think this should be clarified.

trey-stafford: I think this is an improvement: 6234789 

We can always adjust later if we come up with a better scheme though!

seaice_ecdr/constants.py
trey-stafford: c471a6a

seaice_ecdr/initial_daily_ecdr.py
trey-stafford: Hmm, lets keep this in mind and maybe address this more holistically. E.g., should we call the code package `seaice_ecdr` or just `seaice_cdr`? 

I suppose it makes sense to start w/ removing the `e` from code references (maybe there'll be a "very enhanced CDR" in the future, vecdr :laughing: ) . The packaging of the code could always change later. For now, the intent of this project is clearly to produce what we're currently calling the ECDR, but no reason for otherwise modular parts of the code to be restricted to the term.

seaice_ecdr/initial_daily_ecdr.py
trey-stafford: Agreed!

seaice_ecdr/initial_daily_ecdr.py
trey-stafford: :100: 

68deff7

PR #14, opened 2023-10-25
Add 19h 37h to idecdr; do not interpolate pole hole by default for idecdr

sc0tts:


sc0tts:
these commits update the add_temporal_interp branch

PR #20, opened 2023-10-30
Add polehole fill to tiecdr (pass 2) instead of idecdr (pass 1) or final daily (pass 3)

sc0tts:


seaice_ecdr/temporal_composite_daily.py
```diff
@@ -83,6 +85,7 @@ def get_standard_initial_daily_ecdr_filename(
     output_directory="",
 ):
     """Return standard ide file name."""
+    # TODO: Perhaps this function should come from seaice_ecdr, not pm_icecon?
```
trey-stafford, line 88: Agreed. We'll address this with the task to ensure the final filename conforms to our expectations (https://trello.com/c/LzPLN1qo/253-ecdr-file-naming-convention)

seaice_ecdr/temporal_composite_daily.py
```diff
@@ -442,6 +407,40 @@ def temporally_interpolated_ecdr_dataset_for_au_si_tbs(
...
+            logger.info("Filled pole hole")
+            is_pole_filled = (cdr_conc_pole_filled != cdr_conc_pre_polefill) & (
+                ~np.isnan(cdr_conc_pole_filled)
+            )
+            if "spatint_bitmask" in tie_ds.variables.keys():
+                tie_ds["spatint_bitmask"].data[
+                    is_pole_filled
+                ] += tb_spatint_bitmask_map["pole_filled"]
+                logger.info("Updated spatial_interpolation with pole hole value")
+        tie_ds["cdr_conc"].data[0, :, :] = cdr_conc_pole_filled[:, :]
```
trey-stafford, line 442: This should be a part of the above `if` block - otherwise `cdr_conc_pole_filled` will not be defined.

seaice_ecdr/temporal_composite_daily.py
```diff
@@ -442,6 +407,40 @@ def temporally_interpolated_ecdr_dataset_for_au_si_tbs(
...
+        #       NOT be definied here (and in the idecdr code...(!))
+        tb_spatint_bitmask_map = {
+            "v18": 1,
+            "h18": 2,
+            "v23": 4,
+            "v36": 8,
+            "h36": 16,
+            "pole_filled": 32,
+        }
+        if cdr_conc.shape == (896, 608):
```
trey-stafford, line 426:
```suggestion
        if hemisphere == NORTH:
```
Use hemisphere here instead of shape.

seaice_ecdr/temporal_composite_daily.py
trey-stafford: This type of error is more easily caught by re-formulating this `if` block. Instead of checking that the hemisphere is 'north' and then doing pole hole filling, it would be more clear to do something like this:

```
if hemisphere != NORTH:
    # Return the dataset unchanged
    return tie_ds

<logic to fill the pole hole>
```

seaice_ecdr/temporal_composite_daily.py
```diff
@@ -622,4 +625,11 @@ def cli(
     hemisphere = NORTH
     resolution = "12"
 
-    gen_temporal_composite_daily(date, hemisphere, resolution)
+    # gen_temporal_composite_daily(date, hemisphere, resolution)
```
trey-stafford, line 628: Remove commented-out code

trey-stafford (changes requested):

seaice_ecdr/temporal_composite_daily.py
sc0tts: I added a comment to the TODO block point to Trello, though not to a specific link.  (Seems unnecessary to have the specific link be in the git history?)

trey-stafford: In this case, I think you can leave it out, but generally adding context like links to issues is encouraged!

seaice_ecdr/temporal_composite_daily.py
sc0tts: I've rewritten the logic here and added some RuntimeError raising as appropriate.

There are actually two sets of logic being conflated here:  Whether or not to do the pole hole filling at all, and whether or not the grid is psn12.5.  The latter logic is implemented here, though it should be abstracted to pm_icecon when we have a more formal way of specifying "how" to do the pole hole filling.  E.g. "fill-with-mean-of non-polehole-values-surrounding-the-pole-hole"...but with a *much* more concise label(!)

seaice_ecdr/temporal_composite_daily.py
sc0tts: I addressed this with the prior code reworking.  This code is specifically designed for psn12.5, so will not work with all NH grids.  (It reads the psn12.5 polehole mask)

seaice_ecdr/temporal_composite_daily.py
sc0tts: Done

trey-stafford (approved):

PR #21, opened 2023-10-31
My updates for add temporal interp

trey-stafford:


seaice_ecdr/temporal_composite_daily.py
```diff
@@ -82,7 +82,7 @@ def get_standard_initial_daily_ecdr_filename(
     date,
     hemisphere,
     resolution,
-    output_directory="",
```
trey-stafford, line 85: I was finding a file in my project directory after each run and I was having trouble tracking down which test produced it. This made it clear where we weren't passing an arg to `output_directory` (which should always be set!).

seaice_ecdr/temporal_composite_daily.py
```diff
@@ -107,18 +107,23 @@ def read_with_create_initial_daily_ecdr(
     date,
     hemisphere,
     resolution,
-    ide_filepath=None,
+    ide_filepath: Path | None = None,
+    ide_dir: Path | None = None,
```
trey-stafford, line 111: Added `ide_dir` because if an `ide_filepath` isn't given, the standard filename needs to be looked up in a directory.

The logic here would be greatly simplified by dropping the `ide_filepath` and requiring the `ide_dir` (this function would _always_ expect a 'standard' filename). There is a test defined to ensure this function can accept a non-standard filename though. 

Is there ever a reason we want to do this? If there isn't a good reason, lets drop this extra logic & kwarg!

sc0tts: The case I am envisioning is: For running tests during development, we may want to have a long series of days already calculated.  And then only work with the latest day or so in the series.

But this is easily accomplished by using a standard filename in a specified directory.

So, I think it's not a problem to always use standard file names.

seaice_ecdr/temporal_composite_daily.py
sc0tts: Good catch!

sc0tts (approved):
These clarifications about file names and directory specification are welcome improvements.

Approved!

seaice_ecdr/temporal_composite_daily.py
trey-stafford: Got it!

trey-stafford: 500b311

PR #26, opened 2023-11-01
Refactor output paths

trey-stafford:
Refactor handling of output paths into `cdr_data_dir`, which will contain subdirectories with outputs/inputs for various processing stages.

This reduces the amount of 'passing paths around' that needs to happen.

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -30,23 +30,23 @@ jobs:
           # Increase this key to trigger cache invalidation
           cache-environment-key: 1
 
+      # Linter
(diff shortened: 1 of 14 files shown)


trey-stafford:
>     1. Consider referring to the output path as "**e**cdr_" rather than "cdr_

Done!
 
>     2. Do we want to further refactor the several functions for path/dir names into one and have that function use subdirs as appropriate for the specific file path -- eg idecdr, tiecdr, cdecdr, monthly, etc -- that are passed as arguments?

Will at least partially address this with a follow-up PR to use standard filenaming convention

PR #27, opened 2023-11-01
Implement standard filenames for CDR V5

trey-stafford:
Adds functions for generating daily and monthly standard filenames.

Utilizes these functions when creating various intermediate outputs.

E.g.,
```
$ ./scripts/cli.sh  cdecdr --date 2022-03-02 --hemisphere north --resolution 12.5
<output clipped for brevity>
$ tree /share/apps/amsr2-cdr/ecdr_v05r00_outputs/
/share/apps/amsr2-cdr/ecdr_v05r00_outputs/
└── standard
    ├── complete_daily
    │   ├── cdecdr_sic_psn12.5km_20220301_ausi_v05r00.nc
    │   └── cdecdr_sic_psn12.5km_20220302_ausi_v05r00.nc
    ├── initial_daily
    │   ├── idecdr_sic_psn12.5_20220224_ausi_v05r00.nc
    │   ├── idecdr_sic_psn12.5_20220225_ausi_v05r00.nc
    │   ├── idecdr_sic_psn12.5_20220226_ausi_v05r00.nc
    │   ├── idecdr_sic_psn12.5_20220227_ausi_v05r00.nc
    │   ├── idecdr_sic_psn12.5_20220228_ausi_v05r00.nc
    │   ├── idecdr_sic_psn12.5_20220301_ausi_v05r00.nc
    │   ├── idecdr_sic_psn12.5_20220302_ausi_v05r00.nc
    │   ├── idecdr_sic_psn12.5_20220303_ausi_v05r00.nc
    │   ├── idecdr_sic_psn12.5_20220304_ausi_v05r00.nc
    │   ├── idecdr_sic_psn12.5_20220305_ausi_v05r00.nc
    │   ├── idecdr_sic_psn12.5_20220306_ausi_v05r00.nc
    │   └── idecdr_sic_psn12.5_20220307_ausi_v05r00.nc
    └── temporal_interp
        ├── tiecdr_sic_psn12.5_20220301_ausi_v05r00.nc
        └── tiecdr_sic_psn12.5_20220302_ausi_v05r00.nc

4 directories, 16 files
```

PR #28, opened 2023-11-01
Standardize supported satellite strs

trey-stafford:
Follow-up PR for #27 that standardizes the strings we use to identify different satellites supported by the ECDR code.

### DIFF
--- a/seaice_ecdr/_types.py
+++ b/seaice_ecdr/_types.py
@@ -1,3 +1,15 @@
 from typing import Literal
 
+# In kilometers.
 ECDR_SUPPORTED_RESOLUTIONS = Literal["12.5", "25"]
+
+# Supported sats
+SUPPORTED_SAT = Literal[
+    "am2",  # AMSR2
+    "ame",  # AMSRE
+    "F17",  # SSMIS F17
+    "F13",  # SSMI F13
+    "F11",  # SSMI F11
+    "F08",  # SSMI F08
+    "n07",  # Nimus SMMR
+]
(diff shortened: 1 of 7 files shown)


seaice_ecdr/_types.py
```diff
@@ -1,3 +1,15 @@
...
+# Supported sats
+SUPPORTED_SAT = Literal[
+    "am2",  # AMSR2
+    "ame",  # AMSRE
+    "F17",  # SSMIS F17
+    "F13",  # SSMI F13
+    "F11",  # SSMI F11
+    "F08",  # SSMI F08
+    "n07",  # Nimus SMMR
+]
```
sc0tts, line 15: One thing to not is that the satellite is not sufficient for determining the input source.

AU_SI12 provides data on the e2n12.5 and e2s12.5 grids.

0631v1 and 0630v2 can provide AMSR2 (and other) data on e2n25, e2n12.5, e2n6.25, e2n3.125 (and the SH equivalent) polar grids.

0630 also has SIR and GRD grid types, and is often re-sampled from the original resolution to other resolutions.

Those would all be "am2" as the satellite, but ... there are many different sources of the TB and methodologies for getting to a particular grid.

sc0tts (approved):
Definitely approve.

Would like to capture the insight that while NSIDC products have traditionally used "sat" as a way of specifying the source of the input (TB) data, this is no longer sufficient.

seaice_ecdr/_types.py
trey-stafford: yes, I think we'll have to deal with that when we start supporting other platforms. For now, this was setup to fulfil a need for establishing the standard "sat" in the filenames.

PR #30, opened 2023-11-02
Set cdecdr ncattrs

sc0tts:
This PR corrects some variable names, data types, and attributes for the output of the completed daily files.


trey-stafford:
One thing I notice about the daily complete files: the NT and BT raw fields have various parameters for their respective algorithms, including e.g., `bt_tb_data_mask_function`, which gets written with a value of something like `"<function tb_data_mask at 0x7faafed6c550>"`

We should make sure to clean up these attrs!

trey-stafford (approved):
There's still a fair amount of TODOs, but we'll address those separately.

PR #31, opened 2023-11-03
Trey fixup nc encoding

trey-stafford:
xarray docs: https://docs.xarray.dev/en/stable/user-guide/io.html#scaling-and-type-conversions

`scaling` and `add_offset` should be passed to the variable's `encoding` dict when writing to netcdf flie.

### DIFF
--- a/seaice_ecdr/complete_daily_ecdr.py
+++ b/seaice_ecdr/complete_daily_ecdr.py
@@ -5,6 +5,7 @@ import copy
 import datetime as dt
 import sys
 import traceback
+from collections import defaultdict
 from functools import cache
 from pathlib import Path
 from typing import Iterable, cast, get_args
@@ -315,11 +316,26 @@ def write_cde_netcdf(
         if excluded_field in cde_ds.variables.keys():
             cde_ds = cde_ds.drop_vars(excluded_field)
 
-    nc_encoding = {}
+    nc_encoding = defaultdict(dict)
     for varname in cde_ds.variables.keys():
         varname = cast(str, varname)
         if varname not in uncompressed_fields:
             nc_encoding[varname] = {"zlib": True}
+
+        # Move scale_factor and add_offset into nc encoding. Without doing this,
+        # the resulting NC file hsa data that, when read, are not scaled
+        # properly (e.g., 255 is 2.55)
+        if "scale_factor" in cde_ds[varname].attrs.keys():
+            logger.info(f"adding scale factor to {varname}")
+            nc_encoding[varname] = {
+                **nc_encoding[varname],
+                "scale_factor": cde_ds[varname].attrs.pop("scale_factor"),
+            }
+        if "add_offset" in cde_ds[varname].attrs.keys():
+            nc_encoding[varname] = {
+                **nc_encoding[varname],
+                "add_offset": cde_ds[varname].attrs.pop("add_offset"),
+            }
 
     cde_ds.to_netcdf(
         output_filepath,
--- a/seaice_ecdr/complete_daily_ecdr.py
+++ b/seaice_ecdr/complete_daily_ecdr.py
@@ -322,20 +322,12 @@ def write_cde_netcdf(
         if varname not in uncompressed_fields:
             nc_encoding[varname] = {"zlib": True}
 
-        # Move scale_factor and add_offset into nc encoding. Without doing this,
-        # the resulting NC file hsa data that, when read, are not scaled
-        # properly (e.g., 255 is 2.55)
-        if "scale_factor" in cde_ds[varname].attrs.keys():
-            logger.info(f"adding scale factor to {varname}")
-            nc_encoding[varname] = {
-                **nc_encoding[varname],
-                "scale_factor": cde_ds[varname].attrs.pop("scale_factor"),
-            }
-        if "add_offset" in cde_ds[varname].attrs.keys():
-            nc_encoding[varname] = {
-                **nc_encoding[varname],
-                "add_offset": cde_ds[varname].attrs.pop("add_offset"),
-            }
+    # scale flag values
+    # TODO: other variables need tihs adjustment?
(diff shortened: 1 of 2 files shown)


seaice_ecdr/complete_daily_ecdr.py
```diff
@@ -305,12 +316,27 @@ def write_cde_netcdf(
...
+    nc_encoding = defaultdict(dict)
     for varname in cde_ds.variables.keys():
         varname = cast(str, varname)
         if varname not in uncompressed_fields:
             nc_encoding[varname] = {"zlib": True}
 
+        # Move scale_factor and add_offset into nc encoding. Without doing this,
+        # the resulting NC file hsa data that, when read, are not scaled
+        # properly (e.g., 255 is 2.55)
+        if "scale_factor" in cde_ds[varname].attrs.keys():
```
sc0tts, line 328: So..it doesn't look like this works quite as intended.  When I tried it, I get values of 254 and 255 instead of 2.54 and 2.55 ... but I also get values of (say) 27 and 100 for ice conc.  And per the standard variable "sea_ice_area_fraction", the range of decoded values must be 0.0-1.0.

I think the values 2.54 and 2.55 (instead of 254 and 255) are a result of the scale_factor (and add_offset) getting applied to the data because mask_and_scale is True...not a result of scale_factor and add_offset being set as variable attributes instead of in the encoding.

trey-stafford: Hmm, maybe there are a couple of issues here. I see that the variable's valid range is 0-100, so I thought that SIC values of 0-100 were expected. That probably needs to be updated to 0-1 (or 0-2.55 if we want flag values left unmasked).

I think to achieve what we're looking for here, when we apply the scale and offset, we also need to adjust our flag values. E.g., 255 should actually be 25500 before being written out (so that when the data is read, the flags have their intended value:

25500 * 0.01 = 255

trey-stafford:
WIP. We need to figure out if there are any other vars that need a similar fix and apply scaling to SIC & flags in a consistent and clear way.

seaice_ecdr/set_daily_ncattrs.py
```diff
@@ -50,7 +50,7 @@ def finalize_cdecdr_ds(
     # cdr_seaice_conc should be ubyte
     ds["cdr_seaice_conc"] = (
         ("time", "y", "x"),
-        ds["cdr_seaice_conc"].data.astype(np.uint8),
+        ds["cdr_seaice_conc"].data.astype(np.uint16),
```
trey-stafford, line 53: For the flag values to be scaled appropriately when we write the data to nc file, we need to use a dtype that will accomodate the unscaled flag values (e.g., 25000)

trey-stafford: If we want the `dtype` of the variable in the nc file to be `uint8`, we may need to pass that as an encoding option for this variable.

seaice_ecdr/set_daily_ncattrs.py
```diff
@@ -64,7 +64,7 @@ def finalize_cdecdr_ds(
             "flag_meanings": "pole_hole lakes coastal land_mask missing_data",
             "reference": "https://nsidc.org/data/g02202/versions/5/",
             "ancillary_variables": "stdev_of_cdr_seaice_conc qa_of_cdr_seaice_conc",
-            "valid_range": np.array((0, 100), dtype=np.uint8),
+            "valid_range": np.array((0, 255), dtype=np.uint8),
```
trey-stafford, line 67: Expand the valid range to include the flag values.

seaice_ecdr/set_daily_ncattrs.py
sc0tts: So...this issue warrants more discussion.

From a CF-compliant, standard name, standard physical quantity sense, the quantity we are providing in these "concentration" fields is actually "sea_ice_area_fraction" [here, siaf] which is a real value between 0.0 and 1.0 (inclusive). 

Because we only need 1% precision in this data product, this is easily encoded with 101 values -- 0 through 100, inclusive -- which are easily converted to siaf by using the uint8 values 0 through 100 and indicating in the data variable attributes that these should be "unpacked" using a scale_factor of 0.01 and an add_offset of 0.0.

The valid range -- alternatively and equivalently expressed as valid_min and valid_max -- are the "packed" limits of valid values in the as-stored array.  Any value outside of that range is, strictly speaking, considered "invalid" and cannot be used to indicate siaf.

Traditionally -- perhaps not great reason, but also not entirely without reason :-) -- sea ice concentration fields have encoded ancillary information in grid cells that do not have valid siaf values.  E.g. indications of land_mask, lakes, permanent_glacial_ice, ice_shelf, this_grid_is_off_earth, missing_data can be indicated by using values that are outside of the valid range.

Use of those non-valid values is non-standard.  I.e, if you are treating this field as a source of information about sea_ice_area_fraction, *all* flag values should be interpreted in the same way: as no-siaf-information-available here, often numerically indicated as a NaN.

So, we certainly don't want to increase the storage space for these arrays by using a larger-than-necessary data type, eg int16 or float32 or double.  

The "extra" information in these grids should be ignorable by any operation that strictly deals with these fields for their standard_name'd content.  

We provide the additional information as a convenience for users, but they are then responsible for knowing how to handle that information _if they want to do so_.  I.e., they can either read in the values as-scaled, knowing that an original value like "255" will be scaled/offsetted to "2.55", or they can read the data variable from the netCDF file and explicitly bypass the masking/scaling so that they have access to the "raw" (=packed) data.

Actually, only the latter of those two options should be considered reliable.  It is entirely valid for software that reads in packed data to cause all values outside of the valid range to return as np.nan, thus ignoring all/any non-standard interpretations of the data field.

The fields in the code are already "packed" the way we want them: as uint8 fields with valid range 0 to 100 and flag values set outside of that range as appropriate.

trey-stafford:
Declining. We think the code-as-is produces files like v4.

PR #32, opened 2023-11-08
Remove scale factor and offset from nc encoding

trey-stafford:
This was accidentally added in the merge of 88a1249


sc0tts (approved):
Good update here.

PR #37, opened 2023-11-15
Add global attrs to monthly files

trey-stafford:
This PR adds a function to create a dictionary of standard global attributes for daily and monthly netcdf files. Although we aren't producing aggregate files yet, this code has support for those too.

Still some TODOs here (e.g., to support different source datasets/platforms).

### DIFF
--- a/seaice_ecdr/monthly.py
+++ b/seaice_ecdr/monthly.py
@@ -552,14 +552,7 @@ def make_monthly_ds(
 
     # Set global attributes
     monthly_ds_global_attrs = get_global_attrs(
-        # TODO: These should be the start and then end of the month, not of the data. So,
-        # 0:00:00 of 1st of month, and 23:59:59.9999 (or whatever) or last day of month.
-        time_coverage_start=dt.datetime(
-            daily_ds_for_month.year, daily_ds_for_month.month, 1, 0, 0, 0
-        ),
-        time_coverage_end=dt.datetime(
-            daily_ds_for_month.year, daily_ds_for_month.month, 1, 23, 59, 59
-        ),
+        time=monthly_ds.time,
         temporality="monthly",
         aggregate=False,
         source=", ".join([fp.item().name for fp in daily_ds_for_month.filepaths]),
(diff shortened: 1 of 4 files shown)


seaice_ecdr/monthly.py
```diff
@@ -550,17 +551,13 @@ def make_monthly_ds(
...
 
     # Set global attributes
-    # TODO: "source" in the v4 files has the full path to the data on FTP. How should we here?
-    monthly_ds.attrs["source"] = ", ".join(
-        [fp.item().name for fp in daily_ds_for_month.filepaths]
+    monthly_ds_global_attrs = get_global_attrs(
+        time=monthly_ds.time,
+        temporality="monthly",
+        aggregate=False,
+        source=", ".join([fp.item().name for fp in daily_ds_for_month.filepaths]),
```
trey-stafford, line 558: We should discuss this. V4 has the path to the data on FTP, but that's OBE. 

I think it's better to just have filenames here, so that the "source" is separate from the platform hosting the data, but we might want to construct a full path to the "source" data on the NOAA HTTPS server.

seaice_ecdr/set_daily_ncattrs.py
```diff
@@ -383,7 +345,14 @@ def finalize_cdecdr_ds(
...
 
     # Finally, address global attributes
-    new_global_attrs = get_global_attrs()
+    new_global_attrs = get_global_attrs(
+        time=ds.time,
+        temporality="daily",
+        aggregate=False,
+        # TODO: support alternative source datasets. Will be AU_SI12 for AMSR2,
+        # AE_SI12 for AMSR-E, and NSIDC-0001 for SSMIS, SSM/I, and SMMR
+        source="Generated from AU_SI12",
```
trey-stafford, line 354: It might be a good idea to create separate functions for daily & monthly attributes that wrap logic like this and ensure that this string is formated consistently. E.g., `get_global_daily_attrs` would take `source_dataset_id`, which is a kwarg typed with a list of literal strings for AU_SI12, AE_SI12, etc. and would construct the `source` string as something like `f"Generaged from {source_dataset_id}"`.

PR #38, opened 2023-11-15
Harmonize data fields

sc0tts:
This sets variables that should have (time, y, x) dimensions properly from when the fields are first created.

Also, flag values are removed from fields, so that all values outside the valid_range are np.nan.  

Scaling of data variables is handled via the "encoding" of each dataarray variable.


seaice_ecdr/temporal_composite_daily.py
```diff
@@ -817,7 +783,8 @@ def temporally_interpolated_ecdr_dataset_for_au_si_tbs(
     )
 
     # set non-conc values to -1
-    is_non_siconc = np.squeeze(tie_ds["cdr_conc"].data > 100)
+    # is_non_siconc = np.squeeze(tie_ds["cdr_conc"].data > 100)
```
trey-stafford, line 786: Remove?

seaice_ecdr/temporal_composite_daily.py
```diff
@@ -817,7 +783,8 @@ def temporally_interpolated_ecdr_dataset_for_au_si_tbs(
...
+    is_non_siconc = tie_ds["cdr_conc"].data > 100
```
trey-stafford, line 787:
```suggestion
    is_non_siconc = tie_ds["cdr_conc"].data > 1
```

This is fractional SIC.

seaice_ecdr/tests/integration/test_initial_daily_ecdr_generation.py
```diff
@@ -169,3 +169,34 @@ def test_can_drop_fields_from_idecdr_netcdf(
...
+        "raw_nt_seaice_conc",
+        "bt_weather_mask",
+        "nt_weather_mask",
+        "invalid_ice_mask",
+        "spatial_interpolation_flag",
+    )
+    for field_name in expected_tyx_fields:
+        nh_data_shape = sample_idecdr_dataset_nh[field_name].shape
+        try:
+            assert len(nh_data_shape) == 3
```
trey-stafford, line 192: A bare assert here is really all we need - `pytest` will print out an informative message about the error and how the shape differs from what's expected.

If you do want to raise a unique error, it would be easier to avoid the `try-except` block:

```
if not len(nh_data_shape) == 3:
    print('Custom error message)
    raise SystemExit("...")
```

trey-stafford: Note you can also provide a custom erorr message this way:

```
assert len(nh_data_shape) == 3, "your custom error message if assertion is `False`"
```

trey-stafford (changes requested):
Looks good! Only a few things I think we should address. My main concern are some commented-out pieces of code/config. I would like to get those cleaned up if they're no longer needed.

seaice_ecdr/melt.py
```diff
@@ -57,7 +57,7 @@ def melting(
     Expects brightness temperatures in degrees Kelvin.
     """
     is_valid_concentration = np.logical_and(
-        concentrations >= CONCENTRATION_THRESHOLD_PERCENT, concentrations <= 100
+        concentrations >= CONCENTRATION_THRESHOLD_FRACTION, concentrations <= 100
```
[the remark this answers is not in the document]
sc0tts: done.  I actually used a CONSTANT here.

seaice_ecdr/temporal_composite_daily.py
```diff
@@ -327,7 +298,7 @@ def temporally_composite_dataarray(
     linint_run = pdist + ndist
     linint_run[linint_run == 0] = 1  # avoid div by zero
     linint = pconc + pdist * linint_rise / linint_run
-    linint = np.round(linint).astype(np.uint8)
+    # linint = np.round(linint).astype(np.uint8)
```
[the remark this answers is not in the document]
sc0tts: yep

seaice_ecdr/temporal_composite_daily.py
sc0tts: ...and change from 100 to 1!

seaice_ecdr/temporal_composite_daily.py
sc0tts: yep

seaice_ecdr/tests/integration/test_initial_daily_ecdr_generation.py
sc0tts: The try/except was left over from using that to help me catch all the cases.  

Fixed.

Also...good to know about the custom error message!

seaice_ecdr/temporal_composite_daily.py
```diff
@@ -783,8 +782,8 @@ def temporally_interpolated_ecdr_dataset_for_au_si_tbs(
     )
 
     # set non-conc values to -1
-    # is_non_siconc = np.squeeze(tie_ds["cdr_conc"].data > 100)
-    is_non_siconc = tie_ds["cdr_conc"].data > 100
+    # is_non_siconc = tie_ds["cdr_conc"].data > 100
```
trey-stafford, line 785: Remove this?

trey-stafford (approved):
Just one more bit of commented-out code I think we should remove, otherwise looks good & can be merged!

seaice_ecdr/temporal_composite_daily.py
sc0tts: yep

PR #45, opened 2023-11-21
CLI for daily aggregates

trey-stafford:
Draft PR. Still TODO:

* Add lat/lon fields.


seaice_ecdr/daily_aggregate.py
```diff
@@ -0,0 +1,200 @@
...
+    sat: SUPPORTED_SAT,
+) -> None:
+    daily_ds = get_daily_ds_for_year(
+        year=year,
+        ecdr_data_dir=ecdr_data_dir,
+        hemisphere=hemisphere,
+        resolution=resolution,
+    )
+
+    # TODO: add lat/lon fields
```
trey-stafford, line 131: CDR v4 stores lat/lon fields in the ancillary file published alongside the data. We could have the lat/lon field stored in the ancillary file and pull from that here.

Alternatively, compute the lat/lon field from the provided `x` and `y` fields. This would provide us the most long-term flexibility for supporting alternative grids/resolutions.

seaice_ecdr/daily_aggregate.py
```diff
@@ -0,0 +1,200 @@
...
+    *,
+    year: int,
+    ecdr_data_dir: Path,
+    hemisphere: Hemisphere,
+    resolution: ECDR_SUPPORTED_RESOLUTIONS,
+) -> xr.Dataset:
+    """Create an xr.Dataset wtih ECDR complete daily data for a given year.
+
+    The resulting xr.Dataset includes:
+        * `year` attribtue.
```
sc0tts, line 60: attribtue -> attribute

seaice_ecdr/daily_aggregate.py
sc0tts: This is a larger question: the authoritative description of grids.
Maybe use NSIDC-0771 ?

sc0tts (approved):
Looking through the new files, I think this looks good.  It's clean, it's similar to similar-functioning code.

I don't know if we have a check to verify that we can retrieve a specific month ... eg "March 21, 2021" from the daily-aggregate or  "March 2021" from the monthly-aggregate and verify that those are the same fields?  

Looks good.

seaice_ecdr/daily_aggregate.py
trey-stafford: We agreed to, at least initially, use NSIDC-0771.

trey-stafford:
> I don't know if we have a check to verify that we can retrieve a specific month ... eg "March 21, 2021" from the daily-aggregate or "March 2021" from the monthly-aggregate and verify that those are the same fields?

Good suggestion - I added this as a regression test!

PR #47, opened 2023-11-30
Update aggregate filenames to exclude `sat`

trey-stafford:
The daily and monthly aggregate files should not include `sat` in the name.


sc0tts (approved):
Looks good

PR #56, opened 2023-12-13
Housekeeping of /share/apps/

trey-stafford:
* Move comparison/visualization code into `old` dir. Plan to remove this soon. This code hasn't been kept up-to-date. Since it was written primarily to satisfy some visualization requirements for phase 1 of our work, I think we could clean this up.
* Extract dup hard-coded paths to constants
* Utilize `pathlib.Path` where it wasn't previously


sc0tts (approved):
Just a quick fyi, there is a "NASA Team 2" sea ice concentration algorithm that is different than the NT2 land spillover technique.  

Knowing this, my first reaction when I saw the directory "nasateam2_ancillary" was "Are we implementing that algorithm here already?"

Just fyi.

PR #57, opened 2023-12-13
Create ECDR ancillary files

trey-stafford:
Replaces the existing mask/geoinformation files


.github/workflows/test.yml
```diff
@@ -39,7 +39,7 @@ jobs:
         uses: "actions/checkout@v3"
         with:
           repository: "nsidc/pm_icecon"
-          ref: "main"
+          ref: "cleanup-external-data-deps"
```
trey-stafford, line 42: Temporary to get the tests to pass here. There are some breaking changes with the new version of `pm_icecon`.

scripts/gather_masks.py
trey-stafford, line 1: This script responsible for creating the ancillary files. This is meant to be run as a 1-off because it relies on paths to data in `/share/apps` that's likely to change soon.

scripts/surface_geo_masks/create_surface_geo_mask.py
trey-stafford, line 1: This script now just produces outputs that are used by `gather_masks.py` (above). The files produced by this script aren't used directly by any of the `seaice_ecdr` code directly anymore.

seaice_ecdr/initial_daily_ecdr.py
```diff
@@ -355,111 +357,42 @@ def compute_initial_daily_ecdr_dataset(
...
+        month=date.month,
+        resolution=resolution,
     )
 
-    # Encode invalid_ice_mask
-    ecdr_ide_ds["invalid_ice_mask"] = (
-        ("time", "y", "x"),
-        np.expand_dims(bt_fields["invalid_ice_mask"], axis=0),
-        {
-            "grid_mapping": "crs",
```
trey-stafford, line 369: Attributes and encoding are now set by the getters for these fields (e.g., in this case, `get_invalid_ice_mask`.

seaice_ecdr/initial_daily_ecdr.py
```diff
@@ -355,111 +357,42 @@ def compute_initial_daily_ecdr_dataset(
...
         hemisphere=hemisphere,
-        resolution=au_si_resolution_str,
     )
     nt_coefs = NtCoefs(
         nt_tiepoints=nt_params.tiepoints,
         nt_gradient_thresholds=nt_params.gradient_thresholds,
     )
 
-    # Encode NT shoremap field
-    ecdr_ide_ds["shoremap"] = (
```
trey-stafford, line 432: We no longer use the `shoremap` or `minic` fields. This was specific to the old method for doing land spillover correction.

seaice_ecdr/initial_daily_ecdr.py
```diff
@@ -637,82 +570,53 @@ def compute_initial_daily_ecdr_dataset(
...
     cdr_conc[set_to_zero_sic] = 0
 
-    # Apply land spillover corrections
-    # TODO: eventually, we want each of these routines to return a e.g.,
-    #   delta that can be applied to the input concentration
-    #   instead of returning a new conc. Then we would have a
-    #   seprate algorithm for choosing how to apply
-    #   multiple spillover deltas to a given conc field.
-
-    use_only_nt2_spillover = True
```
trey-stafford, line 647: We decided to use the NT2 spillover technique. Remove the OBE code.

seaice_ecdr/masks.py
```diff
@@ -47,6 +58,14 @@ def _get_sat_by_date(
 
 
 def _bitmask_value_for_meaning(*, var: xr.DataArray, meaning: str):
+    index = var.flag_meanings.split(" ").index(meaning)
+    # TODO: should this be `flag_masks`?
+    value = var.flag_values[index]
```
trey-stafford, line 63: Question for @sc0tts : I noticed that the geo/information files you created use `flag_values` for the polehole variable. Should that be `flag_masks` to indicate a bitmask?

sc0tts: Yes, good catch.  These should be flag_masks rather than flag_values.

sc0tts (approved):
This all looks good to me.  Looking forward to running code using a single ancillary file!

seaice_ecdr/masks.py
trey-stafford: Will resolve w/ https://trello.com/c/pNy8TTaw/330-ecdr-ensure-flagmasks-are-used-for-bitmask-fields

PR #60, opened 2023-12-19
Apply land mask to QA/flag fields

trey-stafford:
Note that e.g., the spatial and temporal interpoloation flag values still have values of 0 over land, but that is expected. We don't use a fill/missing value for these fields because we want to ensure they remain as `byte`s when read by other programs.

### DIFF
--- a/seaice_ecdr/initial_daily_ecdr.py
+++ b/seaice_ecdr/initial_daily_ecdr.py
@@ -711,6 +711,8 @@ def compute_initial_daily_ecdr_dataset(
     #  32: Spatial interpolation applied
     #  64: *applied later* Temporal interpolation applied
     # 128: *applied later* Melt onset detected
+    # TODO: dynamically read the bitmask values from the source dataset
+    # (`flag_masks` & `flag_meanings`)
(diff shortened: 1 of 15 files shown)


seaice_ecdr/initial_daily_ecdr.py
```diff
@@ -716,6 +718,8 @@ def compute_initial_daily_ecdr_dataset(
     qa_bitmask[invalid_tb_mask & ~ecdr_ide_ds["invalid_ice_mask"].data[0, :, :]] += 8
```
sc0tts, line 718: Perhaps we could put a "TODO" here so that the values that are getting added are determined from a flag_masks index/value instead of being hardcoded as "8" or other bitmask values?

That's simply for clarity in the future, and to tie the value to the flag meanings/values/masks ... in case those change.

sc0tts (approved):
I think this all looks good.

seaice_ecdr/initial_daily_ecdr.py
trey-stafford: f62c29a

PR #61, opened 2023-12-21
Monthly integration test

trey-stafford:
Adds simple integration tests for daily complete and monthly netcdf files.

The tests just assert that the code for producing NetCDFs doesn't fail. We can build more robust tests from here!

One key addition is `pytest-order`, which allows us to define the order in which tests are executed. The daily complete integration test runs first, producing 3 complete data files in a temporary directory. These data files are then read by the code that produces a monthly file from those data.

Eventually I'd like to update the initial daily integration tests to write data to this temporary dir too. Then we can test each phase and not worry about re-creating data unnecessarily - that takes time!

### DIFF
--- a/conda-lock.yml
+++ b/conda-lock.yml
@@ -13,7 +13,7 @@
 version: 1
 metadata:
   content_hash:
-    linux-64: ee261755637274e93ade8e4b9fef780c5027e73884d9580b96852509abab43d7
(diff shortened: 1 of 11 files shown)


sc0tts (approved):
pytest.mark.order and monkeypatch are new to me.  They look *very* useful!

PR #62, opened 2023-12-21
Rename ancillary dir

trey-stafford:
Make the ancillary dir name consistent w/ the base output dir (include the entire version string).

### DIFF
--- a/seaice_ecdr/constants.py
+++ b/seaice_ecdr/constants.py
@@ -24,4 +24,4 @@ NRT_BASE_OUTPUT_DIR = BASE_OUTPUT_DIR / "nrt"
 LANCE_NRT_DATA_DIR = NSIDC_NFS_SHARE_DIR / "lance_amsr2_nrt_data"
 
 # Location of surface mask & geo-information files.
-CDR_ANCILLARY_DIR = NSIDC_NFS_SHARE_DIR / "cdrv5_ancillary"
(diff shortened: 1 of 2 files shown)


seaice_ecdr/constants.py
```diff
@@ -24,4 +24,4 @@
 LANCE_NRT_DATA_DIR = NSIDC_NFS_SHARE_DIR / "lance_amsr2_nrt_data"
 
 # Location of surface mask & geo-information files.
-CDR_ANCILLARY_DIR = NSIDC_NFS_SHARE_DIR / "cdrv5_ancillary"
+CDR_ANCILLARY_DIR = NSIDC_NFS_SHARE_DIR / f"ecdr_{ECDR_PRODUCT_VERSION}_ancillary"
```
sc0tts, line 27: Since we're standardizing file names, should we cause this file name to link "more precisely" to the data product?  E.g.:  "G02202_v5_ancillary/"

trey-stafford: `NSIDC_NFS_SHARE_DIR` will be updated to include `G02202` in the name, so I don't think we need to repeat that here.

trey-stafford:
> Perhaps we want to remove reference to "eCDR" in "official" directories and filenames?

Yes, good point. I'll drop the `eCDR` from these dirs!

PR #64, opened 2024-01-03
Tweaks for run of entire AMSR2 record

sc0tts:
This PR adds tweaks to the code to allow all dates in the AMSR2 record to run.

### DIFF
--- a/seaice_ecdr/ancillary.py
+++ b/seaice_ecdr/ancillary.py
@@ -55,8 +55,16 @@ def _get_sat_by_date(
     # TODO: these date ranges belong in a config location
     if date >= dt.date(2012, 7, 2) and date <= dt.date(2030, 12, 31):
         return "am2"
+    # TODO: Remove this logic when SSMIS satellites are used.
     elif date >= dt.date(2012, 6, 27) and date <= dt.date(2012, 7, 1):
-        logger.warning(f"WARNING: Using AMSR2 pole hole for date: {date}")
+        logger.warning(
+            "_get_sat_by_date() override implemented so that"
+            f" AMSR2 is considered the 'satellite for this date' ({date})"
+            "while this product is under development.  When another"
+            " SSMIS satellite (eg F17 or F18) is used through 7/1/2012,"
+            " this override should be removed.  This call was needed for"
+            " to determine the pole hole mask for this date."
+        )
         return "am2"
     else:
         raise RuntimeError(f"Could not determine sat for date: {date}")
--- a/seaice_ecdr/ancillary.py
+++ b/seaice_ecdr/ancillary.py
@@ -12,6 +12,7 @@ from pathlib import Path
 import numpy as np
 import pandas as pd
 import xarray as xr
+from loguru import logger
 from pm_tb_data._types import NORTH, Hemisphere
 
 from seaice_ecdr._types import ECDR_SUPPORTED_RESOLUTIONS, SUPPORTED_SAT
@@ -53,6 +54,9 @@ def _get_sat_by_date(
     """Return the satellite used for this date."""
     # TODO: these date ranges belong in a config location
     if date >= dt.date(2012, 7, 2) and date <= dt.date(2030, 12, 31):
+        return "am2"
+    elif date >= dt.date(2012, 6, 27) and date <= dt.date(2012, 7, 1):
+        logger.warning(f"WARNING: Using AMSR2 pole hole for date: {date}")
         return "am2"
     else:
         raise RuntimeError(f"Could not determine sat for date: {date}")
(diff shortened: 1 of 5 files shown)


seaice_ecdr/ancillary.py
```diff
@@ -54,6 +55,9 @@ def _get_sat_by_date(
     # TODO: these date ranges belong in a config location
     if date >= dt.date(2012, 7, 2) and date <= dt.date(2030, 12, 31):
         return "am2"
+    elif date >= dt.date(2012, 6, 27) and date <= dt.date(2012, 7, 1):
+        logger.warning(f"WARNING: Using AMSR2 pole hole for date: {date}")
```
trey-stafford, line 59: This function doesn't have anything to do with the pole hole. Change the warning message.

Is this temporary code or do we actually want to use `am2` for this date range long-term?

seaice_ecdr/complete_daily_ecdr.py
```diff
@@ -202,6 +202,15 @@ def create_melt_onset_field(
             dtype=np.uint8,
         )
         logger.info(f"using empty melt_onset_field for prior for {day_of_year}")
+    elif (date.year == 2012) and (date <= dt.date(2012, 7, 2)):
```
trey-stafford, line 205: To confirm: this is because there isn't any data available for AMSR2? When we extend support for other sats this code will change to create the melt field from actual data? 

If so, lets add a comment here explaining.

seaice_ecdr/complete_daily_ecdr.py
```diff
@@ -274,11 +283,29 @@ def complete_daily_ecdr_dataset_for_au_si_tbs(
...
+        name="surface_type",
+        data=surfacetype_da.data,
+        dims=["time", "y", "x"],
+        coords=dict(
+            time=cde_ds.time,
+            y=cde_ds.variables["y"],
+            x=cde_ds.variables["x"],
+        ),
+        attrs=surfacetype_da.attrs,
+    )
```
trey-stafford, line 308: Lets change this to `cde_ds = cde_ds.merge(surf_da.rename("surface_type"))`, which accomplishes the same without needing to re-create a `DataArray`.

seaice_ecdr/complete_daily_ecdr.py
```diff
@@ -379,6 +406,7 @@ def make_cdecdr_netcdf(
         ecdr_data_dir=ecdr_data_dir,
     )
 
+    # TODO: Here, there are no x var attrs
```
trey-stafford, line 409: Still TODO, or can we remove?

seaice_ecdr/initial_daily_ecdr.py
```diff
@@ -741,6 +741,99 @@ def compute_initial_daily_ecdr_dataset(
     return ecdr_ide_ds
 
 
+def create_null_au_si_tbs(
```
trey-stafford, line 744: This is fine for now, but we might consider moving this into the `pm_tb_data` package?

Will re-think this as we begin to support other satellites.

trey-stafford (changes requested):
Looks good - just a few things to address first!

seaice_ecdr/ancillary.py
sc0tts: Actually, this message is here because the function is being called for the pole hole.  I've updated the warning message.

seaice_ecdr/complete_daily_ecdr.py
sc0tts: Correct.  I've added a detailed "TODO:" before this code.

seaice_ecdr/complete_daily_ecdr.py
sc0tts: Good idea.  I'll do that in two steps:  First, re-assign the coordinates of the surfacetype dataarray to use the x, y, time variables from the cde_ds, then merge that dataarray into the cde_ds.

The rename() is no longer necessary because the name is properly set in the ancillary.py code

seaice_ecdr/complete_daily_ecdr.py
sc0tts: Good catch.  Removed.

seaice_ecdr/initial_daily_ecdr.py
sc0tts: I agree.  I decided to put it here because:
(a) I would not have to modify the pm_tb_data repository, and 
(b) seaice_ecdr knows what it "wants" from pm_tb_data...specifically what set of TB arrays it wants.  It seemed clearest to explicitly specify those here (in this repository) rather than assume that all users of pm_tb_data would want the same (null) set of TBs.

There is probably a use case for requesting a specific block of information from the pm_tb_data package -- sat, resolution, grid, dates, etc -- but...I don't think we want to go down that rabbit hole yet.

seaice_ecdr/complete_daily_ecdr.py
trey-stafford: Awesome, thanks!

trey-stafford (approved):

PR #66, opened 2024-01-04
Add surface type mask var to monthly outputs

trey-stafford:
Aggregate files  have a time-varying surface type mask to support different pole hole sizes.


sc0tts (commented):
This looks good to me.  Always good to see tests associated with new code!

sc0tts (approved):

PR #67, opened 2024-01-07
Generate 25km ancillary files

sc0tts:
This PR adds a script named .../seaice_ecdr/scripts/adapt_cdrv5_anc.py
which allows the creation of 25km polar stereo ancillary files given:
  - 12.5km polar stereo ancillary files
  -  nsidc0780 files (region and surface-type data sets, including polar stereo)
  -  nsidc0771 files (polar stereo lat/lon data set)

Usage:
   python adapt_cdrv5_anc.py psn25
   python adapt_cdrv5_anc.py pss25

NOTE: this will not OVERWRITE the existing 25km ancillary file at:
   /share/apps/G02202_V5/v05r00_ancillary/ecdr-ancillary-psn25.nc
   /share/apps/G02202_V5/v05r00_ancillary/ecdr-ancillary-pss25.nc

### DIFF
--- a/scripts/adapt_cdrv5_anc.py
+++ b/scripts/adapt_cdrv5_anc.py
@@ -20,17 +20,19 @@ Specifically:
            latitude: latlon
           longitude: latlon
        surface_type: regions
+             adj123: <code>
   min_concentration: ecdr_anc
    invalid_ice_mask: ecdr_anc
    polehole_bitmask: ecdr_anc
                l90c: <external code>
-             adj123: <external code>
 """
 
 import datetime as dt
 
 import numpy as np
 import xarray as xr
+from pm_icecon.land_spillover import create_land90
+from scipy.signal import convolve2d
 
 ecdr_anc_fn = {
     "psn12.5": "/share/apps/G02202_V5/v05r00_ancillary/ecdr-ancillary-psn12.5.nc",
@@ -56,8 +58,6 @@ nsidc0780_fn = {
 
 def find_coast(mask, ocean_values=(50,), land_values=(75, 250)):
     """Find indices where land values are adjacent to ocean values."""
-    from scipy.signal import convolve2d
-
     is_ocean = np.zeros(mask.shape, dtype=bool)
     for val in ocean_values:
         is_ocean[mask == val] = True
@@ -75,8 +75,13 @@ def find_coast(mask, ocean_values=(50,), land_values=(75, 250)):
     return is_coast
 
 
-def calc_surfacetype_da(ds):
+def calc_surfacetype_np(ds):
     """Calculate the DataArray for the CDRv5 surface_type given a regions DS.
+
+    Note: Input is a Dataset that we can test for various surfacetype field
+          names.
+          Output is a numpy array, so that the data can be assigned to a
+          DataArray with the overall Datasets coord variables.
 
     For the CDRv5 field, the ancillary surface types are:
         50: ocean
@@ -121,7 +126,56 @@ def calc_surfacetype_da(ds):
         land_values=(75, 250),
     )
     surface_mask[is_coast] = 200
-    breakpoint()  # here
+
+    return surface_mask
+
+
+def calc_adj123_np(surftype_da, ocean_val=50, coast_val=200):
+    """Compute the land-adjacency field for this surfacetype mask.
+    Input:
+        DataArray with values:
+            ocean: ocean_val (default 50)
+            coast: coast_val (default 200)
+    Output:
+        Numpy array with adjacency values of 1, 2, 3
+    """
+    surftype = surftype_da.as_numpy()
+    is_ocean = surftype == ocean_val
+    is_coast = surftype == coast_val
+    is_land = (~is_ocean) & (~is_coast)
+
+    kernel = [
+        [
+            1,
+            1,
+            1,
+        ],
+        [1, 1, 1],
+        [
+            1,
+            1,
+            1,
+        ],
+    ]
+    adj123_arr = np.zeros(surftype.shape, dtype=np.uint8)
+    adj123_arr[is_land] = 255
+    for adj_val in range(1, 4):
+        is_unlabeled = adj123_arr == 0
+        is_labeled = (adj123_arr == 255) | ((~is_unlabeled) & (adj123_arr < adj_val))
+        convolved = convolve2d(
+            is_labeled.astype(np.uint8),
+            kernel,
+            mode="same",
+            boundary="fill",
+            fillvalue=0,
+        )
+        is_newly_labeled = is_unlabeled & (convolved > 0)
+        adj123_arr[is_newly_labeled] = adj_val
+
+    # Remove the land grid cells from the adjacency matrix
+    adj123_arr[adj123_arr == 255] = 0
+
+    return adj123_arr
 
 
 def generate_ecdr_anc_file(gridid):
@@ -155,7 +209,13 @@ def generate_ecdr_anc_file(gridid):
             longitude=ds_latlon.variables["longitude"],
         ),
         coords=dict(
-            month=(["month"], months),
+            month=(
+                ["month"],
+                months,
+                {
+                    "long_name": "month of the year",
+                },
+            ),
             x=ds_latlon.variables["x"],
             y=ds_latlon.variables["y"],
         ),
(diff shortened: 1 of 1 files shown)


scripts/adapt_cdrv5_anc.py
```diff
@@ -0,0 +1,575 @@
...
+        coords=dict(
+            y=ds.variables["y"],
+            x=ds.variables["x"],
+        ),
+        attrs=dict(
+            short_name="l90c",
+            long_name=f"{gridid}_land-as-90-percent-concentration_field",
+            grid_mapping="crs",
+            comment="The 'land90' array is a mock sea ice concentration"
+            " array that is calculatedfrom the land mask.  It assumes that"
```
trey-stafford, line 430:
```suggestion
            " array that is calculated from the land mask.  It assumes that"
```

scripts/adapt_cdrv5_anc.py
```diff
@@ -0,0 +1,575 @@
...
+            " the mock concentration value will bethe average of a 7x7 array"
+            " of local surface mask values centered on thecenter pixel."
+            "  Water grid cells are considered to have a sea iceconcentration"
```
trey-stafford, line 433:
```suggestion
            "  Water grid cells are considered to have a sea ice concentration"
```

scripts/adapt_cdrv5_anc.py
```diff
@@ -0,0 +1,575 @@
...
+            " of zero.  Land grid cells are considered to have a sea"
+            " iceconcentration of 90%.  The average of the 49 grid cells"
+            " in the 7x7 arrayyields the `land90` concentration value.",
```
trey-stafford, line 436: A few concatenation errors in this paragraph that need to be fixed. E.g., 

> "array that is **calculatedfrom** the land mask"

scripts/adapt_cdrv5_anc.py
```diff
@@ -0,0 +1,575 @@
...
+    # Set up a new xarray Dataset for the new ancillary file
+    # with values from an authoritative geolocation source
+    # Coords added: month, x, y
+    # Variables added: crs
+    # Attributes added: description and geo-informative attrs
+    ds_latlon = xr.load_dataset(latlon_fn)
+    months = np.arange(1, 13).astype(np.int16)
+    ds = xr.Dataset(
+        data_vars=dict(
+            crs=ds_latlon.variables["crs"],
```
trey-stafford, line 335: One thing I notice in the resulting netcdf is that the `crs` variable is of `int` type and has a scalar value of `-2147483647`. This is present in the 12.5km ancillary files' `crs` variable as well.

Conventionally this variable is of type `char` and doesn't take on a value, right? Does it matter? E.g., `gdal` & `QGIS` seem to understand the CRS without any problems.

scripts/adapt_cdrv5_anc.py
```diff
@@ -0,0 +1,575 @@
...
+            latitude=ds_latlon.variables["latitude"],
```
trey-stafford, line 336: These lat/lon variables have a `missing_value` attribute, but the 12.5km ancillary files do not. Should they?

scripts/adapt_cdrv5_anc.py
```diff
@@ -0,0 +1,575 @@
...
+        data=invalid_ice_mask_np,
+        dims=["month", "y", "x"],
+        coords=dict(
+            month=ds.variables["month"],
+            y=ds.variables["y"],
+            x=ds.variables["x"],
+        ),
+        attrs=dict(
+            short_name="invalid_ice_mask",
+            long_name=f"{gridid} invalid sea ice mask",
```
trey-stafford, line 500: The 12.5km ancillary files don't have a `short_name` attribute for `invalid_ice_mask`.

The 12.5km ancillary files also don't currently include the `gridid` in the `long_name`. 

Should we update the 12.5km ancillary files to be consistent w/ the 25km ancillary files?

scripts/adapt_cdrv5_anc.py
```diff
@@ -0,0 +1,575 @@
...
+        coords=dict(
+            y=ds.variables["y"],
+            x=ds.variables["x"],
+        ),
+        attrs=dict(
+            short_name="adj123",
+            long_name=f"{gridid}_adjacency_field",
+            grid_mapping="crs",
+            flag_values=np.uint8((0, 1, 2, 3)),
+            flag_meanings="not_near_land one_gridcell_from_land two_gridcells_from_land three_gridcell_from_land",
```
trey-stafford, line 411: The 12.5km ancillary files don't have these additional attributes (just a `comment` that indicates "Diagonal adjacency 123 field". We should probably add these attrs to the `12km` ancillary files' `adj123` variable!

scripts/adapt_cdrv5_anc.py
```diff
@@ -0,0 +1,575 @@
...
+        ),
+        attrs=dict(
+            short_name="min_concentration",
+            long_name=f"{gridid} minimum observed NASATeam seaice concentration",
+            grid_mapping="crs",
+            comment="The 'min_concentration' array is the minimum sea ice"
+            " concentration value observed at near-land pixels during"
+            " no-sea-ice conditions.  It represents the amount of apparent"
+            " sea ice concentration attributable to land spillover effect.",
+        ),
```
trey-stafford, line 477: Again, the 12.5km ancillary files don't have these additional attributes for `min_concentration`.

trey-stafford (changes requested):
Mostly looks good! 

The only thing we should definitely address before merging this is the comment related to some concatenation errors in an attribute string.

Most of the other comments are related to inconsistencies between the 25km ancillary files and the 12.5km ancillary files. We should probably create a separate card (edit - [here it is](https://trello.com/c/GLjawK55/338-ecdr-ensure-125km-and-25km-ancillary-files-have-consistent-variable-attributes)) to ensure consistency (e.g., the 25km ancillary files have additional attributes for variables like `min_concentration` that are not in the 12.5km ancillary files.

Shouldn't affect processing, but we do plan to publish these ancillary datasets!

trey-stafford:
Oh, and one other suggestion might be to create a new `ancillary` subdir in `scripts` that we can put this and the 12.5km ancillary file generation script into. That'll help keep `scripts/` organized a little better.

scripts/adapt_cdrv5_anc.py
sc0tts: Good catch.  Fixed the several wordmashings.

scripts/adapt_cdrv5_anc.py
sc0tts: I don't know if there's supposed to be a standard vartype -- and value? -- for the "crs" variable.  The variable is only used as a place to store information in the form of attributes.  Neither the value nor even the data type of the variable itself is ever used.  (Or, more precisely, I've never seen it used anywhere.)

scripts/adapt_cdrv5_anc.py
sc0tts: Good catch.  "missing_value" is deprecated and "_FillValue" should be used.  Though..."_FillValue" should not be used for (a) coordinate/dimension variables, nor for (b) fields -- such as flag or bit masks -- that should not be converted to float upon reading.

I've removed "missing_value" from the latitude and longitude data_vars, though I have left "_FillValue" assigned.

scripts/adapt_cdrv5_anc.py
sc0tts: I've added TODO: comments above the attrs= assignments for each of the fields in the ancillary file.

Before publication, these should be standardized and consistent across the ancillary files for all grids.

scripts/adapt_cdrv5_anc.py
sc0tts: Agree.  We will standardize the field's attributes before publication.

scripts/adapt_cdrv5_anc.py
sc0tts: Agree.  We will standardize the field's attributes before publication.

sc0tts:
> Oh, and one other suggestion might be to create a new `ancillary` subdir in `scripts` that we can put this and the 12.5km ancillary file generation script into. That'll help keep `scripts/` organized a little better.

Good idea.  Perhaps even "create_ancillary/" as the subdir name?

trey-stafford (approved):

PR #69, opened 2024-01-10
Fixes for AMSR2 monthly/aggregate runs

trey-stafford:
* Fix issue with aggregate outputs being larger than expected and taking a long time to create
* Remove `seaice` dependency and `old/` dir in project root containing OBE comparison/visualization code. This allows the `netCDF4` library to upgrade, resulting fixing error messages related to `xarray` and `dask` writing netcdf files to disk.
* Add `--end-year` option for daily aggregate CLI
* Make `time` an unlimited dimension in output aggregate files


environment.yml
```diff
@@ -14,8 +14,7 @@ dependencies:
   - dask ~=2023.10.1
   - netCDF4 ~=1.6.3
   - loguru ~=0.7
-  - scipy ~=1.8.1 # TODO: we should be able to upgrade (.eg., ~=1.11.3) but `seaice` depends on an earlier version. `seaice` isn't _really_ a dependency.
-  - seaice ~=6.0
```
trey-stafford, line 18: Removed `seaice` dependency, which allowed the `netCDF4` library to update. This was important because the version we were on was raising warnings/errors when using `xarray` with `dask`.

sc0tts: Ugh, I've noticed this un-upgradedness of seaice before.  I think we included it for some early plots/comparisons?
Good catch!

seaice_ecdr/monthly_aggregate.py
```diff
@@ -52,6 +53,7 @@ def get_monthly_aggregate_ds(
         resolution=resolution,
     )
 
+    _c = Client(n_workers=2, threads_per_worker=1)
```
sc0tts, line 56: You'll have to tell me how to work with this type of dask assignment...:-)

seaice_ecdr/monthly.py
```diff
@@ -475,20 +475,29 @@ def _assign_time_to_monthly_ds(
...
+    hemisphere: Hemisphere,
 ) -> xr.DataArray:
-    daily_surf_mask = daily_ds_for_month.surface_type_mask
+    try:
+        daily_surf_mask = daily_ds_for_month.surface_type_mask
+    except AttributeError:
+        logger.warning(
+            f"Daily data for {daily_ds_for_month.year}-{daily_ds_for_month.month} use old `surface_type` varname."
+        )
+        daily_surf_mask = daily_ds_for_month.surface_type
```
trey-stafford, line 486: TODO: remove this try/except behavior before merging. We'll re-process  the daily data.

environment.yml
trey-stafford: Yes - TODO: remove `old/` code that contains references to the `seaice` library.

trey-stafford:
> We may yet need to simplify the surface_type fields. I'm not sure why they are increasing the size of the final netCDF file by...a factor of 12 or more(!)

This issue resolved by utlizing `ncrcat` instead of `xarray` to do the concatenation of individual daily/monthly files.

environment.yml
```diff
@@ -12,10 +12,11 @@ dependencies:
   - numpy ~=1.22.4
   - xarray ~=2023.6
   - dask ~=2023.10.1
-  - netCDF4 ~=1.6.3
+  - netCDF4 ~=1.6.5
   - loguru ~=0.7
-  - scipy ~=1.8.1 # TODO: we should be able to upgrade (.eg., ~=1.11.3) but `seaice` depends on an earlier version. `seaice` isn't _really_ a dependency.
-  - seaice ~=6.0
+  - scipy ~=1.8.1
```
sc0tts, line 17: Do we want to try increasing this scipy version back up to 1.11.3 or so now that the seaice dependency has been removed?

sc0tts (approved):
Looks good to me.
Very interesting how much faster nco is than xarray....

PR #70, opened 2024-01-10
Fix missing x y attrs sh

trey-stafford:


scripts/ancillary/fix-sh-xy-attrs.py
```diff
@@ -0,0 +1,50 @@
...
+from seaice_ecdr.ancillary import get_ancillary_filepath
+
+
+def fix_xy_attrs(*, ancillary_ds: xr.Dataset) -> xr.Dataset:
+    ancillary_ds.x.attrs = dict(
+        standard_name="projection_x_coordinate",
+        long_name="x coordinate of projection",
+        units="meters",
+        axis="X",
+        valid_range=(ancillary_ds.x.min(), ancillary_ds.x.max()),
```
sc0tts, line 19: I think this valid range assignment isn't quite right.
There's a difference between the "valid_range" and the "actual_range".  

The latter refers to the values in the arrays -- i.e., what would be returned by .min() and .max() functions.

The valid_range should include the entire domain of the grid, and so should go from "x.min() - half-the-grid-cell-width" to "x.max() + half-the-grid-cell-width"

scripts/ancillary/fix-sh-xy-attrs.py
```diff
@@ -0,0 +1,50 @@
...
+    )
+
+    ancillary_ds.y.attrs = dict(
+        standard_name="projection_y_coordinate",
+        long_name="y coordinate of projection",
+        units="meters",
+        axis="Y",
+        valid_range=(ancillary_ds.y.min(), ancillary_ds.y.max()),
```
sc0tts, line 27: As above, valid_range != actual_range

sc0tts (changes requested):
At some point, we probably need to have a single-source-of-truth for the geospatial netCDF variables/coordinates/dimensions.  

For now, I've been using the appropriate fields in NSIDC-0771 for the polar-stereo grids.

scripts/ancillary/fix-sh-xy-attrs.py
trey-stafford: Good catch!

trey-stafford: 38c2115

PR #72, opened 2024-01-17
Simplify platform start date overrides

trey-stafford:
Override platform start dates via a config file and envvar. This approach can also be extended to other overrides you might want to apply to other configuration. For now, this PR just adds support for overriding the platform start dates.

The key change this PR makes is setting up a single place for the code that handles overrides to operate. At runtime, we check if an overrides file is specified. If so, we read start dates from that config file. Otherwise, we use the defaults as defined in our code.

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -69,6 +69,6 @@ module = [
   "setuptools.*",
   "netCDF4.*",
   "rasterio.*",
-  "seaice.*",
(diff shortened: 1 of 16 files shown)


sc0tts (approved):
Excellent!
I tested this against my ridiculously-verbose version and got identical results.
This is a great way to pass information down deep in the code.

PR #75, opened 2024-01-18
Read f17 from 0001

trey-stafford:


seaice_ecdr/ancillary.py
```diff
@@ -38,7 +38,10 @@ def get_ancillary_ds(
     *, hemisphere: Hemisphere, resolution: ECDR_SUPPORTED_RESOLUTIONS
 ) -> xr.Dataset:
     """Return xr Dataset of ancillary data for this hemisphere/resolution."""
-    if resolution != "12.5":
+    # TODO: This list could be determined from an examination of
+    #       the ancillary directory
+    implemented_resolutions = ("12.5", "25")
```
trey-stafford, line 43: We should use `seaice_ecdr._types.ECDR_SUPPORTED_RESOLUTIONS` here.

I'm not sure that the existence of ancillary files alone is enough. That would be nice, but I think there's still a fair bit of code that relies on hard-coded resolution checks.

seaice_ecdr/ancillary.py
```diff
@@ -38,7 +38,10 @@ def get_ancillary_ds(
...
+    if resolution not in implemented_resolutions:
         raise NotImplementedError("ECDR currently only supports 12.5km resolution.")
```
trey-stafford, line 45: Error message should be updated to indicate the expected resolutions.

seaice_ecdr/ancillary.py
```diff
@@ -71,7 +74,15 @@ def _get_sat_by_date(
...
 def bitmask_value_for_meaning(*, var: xr.DataArray, meaning: str):
-    index = var.flag_meanings.split(" ").index(meaning)
+    # TODO: where do we encounter the ValueError that requires setting
+    # `meaning.lower()`? Can we just use a `.lower()` on the `flag_meanings` and
+    # `meaning` to be consistent and never run into this problem? Are there any
+    # cases where case matters?
+    try:
+        index = var.flag_meanings.split(" ").index(meaning)
+    except ValueError:
+        index = var.flag_meanings.split(" ").index(meaning.lower())
```
trey-stafford, line 84: Curious where the need for this came from. Do we have some inconsistent flag_meanings strings? 

Would it be safe to just do this instead?

```
index = var.flag_meanings.lower().split(" ").index(meaning.lower())
```

^that would ensure both strings are always lower-case. That said, I feel like the need for this in the first place is flag that something else is wrong in our code.

seaice_ecdr/grid_id.py
```diff
@@ -4,7 +4,8 @@
 
 from seaice_ecdr._types import ECDR_SUPPORTED_RESOLUTIONS
 
-GRID_ID = Literal["psn12.5", "pss12.5"]
+# GRID_ID = Literal["psn12.5", "pss12.5"]
```
trey-stafford, line 7: Remove commented out `GRID_ID`.

seaice_ecdr/initial_daily_ecdr.py
```diff
@@ -34,7 +35,10 @@
 from pm_tb_data.fetch.amsr.ae_si import get_ae_si_tbs_from_disk
 from pm_tb_data.fetch.amsr.au_si import get_au_si_tbs
 from pm_tb_data.fetch.amsr.util import AMSR_RESOLUTIONS
+from pm_tb_data.fetch.nsidc_0001 import NSIDC_0001_SATS, get_nsidc_0001_tbs_from_disk
 
+# TODO: default flag values are specific to the ECDR, and should probably be
+# defined in this repo instead of `pm_icecon`.
```
trey-stafford, line 41: This TODO should be moved to the associated import of `DEFUALT_FLAG_VALUES` from `pm_icecon`. 

I think this is OBE too...we could probably just remove this import and use a fixed value of e.g., 255. This is only used to get the `missing_value` that's used in the bootstrap calculation.. It then gets overwritten (the flag value never ends up in the CDR's outputs).

seaice_ecdr/initial_daily_ecdr.py
```diff
@@ -337,7 +350,10 @@ def compute_initial_daily_ecdr_dataset(
             != ecdr_ide_ds[si_varname].data[0, :, :]
         ) & (~np.isnan(ecdr_ide_ds[si_varname].data[0, :, :]))
         spatint_bitmask_arr[is_tb_si_diff] += TB_SPATINT_BITMASK_MAP[tbname]
-        land_mask = get_land_mask(hemisphere=hemisphere, resolution=resolution)
+        land_mask = get_land_mask(
+            hemisphere=cast(Hemisphere, hemisphere),
```
trey-stafford, line 354: this cast should be unnecessary. Remove.

seaice_ecdr/initial_daily_ecdr.py
```diff
@@ -365,11 +381,28 @@ def compute_initial_daily_ecdr_dataset(
...
+            gridid=ecdr_ide_ds.grid_id,
+        )
+    elif platform == "ame":
+        bt_coefs_init = pmi_bt_params_amsre.get_ausi_amsre_bootstrap_params(
+            date=date,
+            satellite="amsre",
+            gridid=ecdr_ide_ds.grid_id,
+        )
+    elif platform == "F17":
+        # TODO: This needs to become ame bootstrap params
```
trey-stafford, line 398: OBE comment?

seaice_ecdr/initial_daily_ecdr.py
```diff
@@ -396,18 +429,33 @@ def compute_initial_daily_ecdr_dataset(
...
             sat=platform,
         )
         ecdr_ide_ds["pole_mask"] = pole_mask
 
     # Determine the NT fields and coefficients
-    nt_params = nt_amsr2_params.get_amsr2_params(
-        hemisphere=hemisphere,
-    )
+    # TODO: Add the rest of these satellites...
+    valid_sats_0001_nt = ("F17",)
```
trey-stafford, line 441: This is defined in `pm_tb_data.fetch.nsidc0001.NSIDC_0001_SATS`.

seaice_ecdr/initial_daily_ecdr.py
```diff
@@ -759,10 +807,39 @@ def compute_initial_daily_ecdr_dataset(
         },
     )
 
-    # Finished!
     return ecdr_ide_ds
 
 
+# TODO: Eventually, these should be renamed based on microwave
+#       band names, not AMSR2 TB channel frequencies
```
trey-stafford, line 814: I would consider this to be a pretty high priority. This is quite confusing unless you understand the history of this code, and I can easily see it tripping us up in the future.

We started with AMSR2, wrote code specific to its TB channels. The AMSR2 channels do not map directly to the channels expected by `pm_icecon`. So the AMSR2 channels get re-mapped/re-named at that step.

This introduces an _additional_ level of misdirection. Now we're mapping Tb channels from other satellites to AMSR2, and then those get mapped again to different channel names for `pm_icecon` as described above.

We should wait to merge w/ main until this task is resolved.

seaice_ecdr/platforms.py
```diff
@@ -107,9 +107,9 @@ def get_platform_start_dates() -> OrderedDict[dt.date, SUPPORTED_SAT]:
                 dt.date(1987, 7, 10): "F08",
                 dt.date(1991, 12, 3): "F11",
                 dt.date(1995, 10, 1): "F13",
-                dt.date(2008, 1, 1): "F17",
-                # dt.date(2002, 6, 1): 'ame',   # AMSR-E...
-                # dt.date(2011, 10, 4): 'F17',  # followed by f17 (again)
+                dt.date(2002, 6, 1): "ame",  # AMSR-E is first AMSR sat
+                # dt.date(2008, 1, 1): "F17", # F17 starts while AMSR-E is up
```
trey-stafford, line 111: Remove OBE line?

seaice_ecdr/regrid_25to12.py
```diff
@@ -0,0 +1,385 @@
...
+#   sudo apt install libgl1
+# Also, the opencv-pytypes package is only available from pip3, not from mamba
+#   hence the type-ignore here
+from cv2 import INTER_LINEAR, resize  # type: ignore[import-not-found]
+from loguru import logger
+from pm_icecon.constants import DEFAULT_FLAG_VALUES
+from pm_tb_data._types import Hemisphere
+
+# TODO: default flag values are specific to the ECDR, and should probably be
+# defined in this repo instead of `pm_icecon`.
```
trey-stafford, line 22: This should be associated with the import for `DEFAULT_FLAG_VALUES`

seaice_ecdr/regrid_25to12.py
```diff
@@ -0,0 +1,385 @@
+"""Routines for regridding 25km polar stereo files to 12.5km
+
+"""
```
trey-stafford, line 3: It would be useful to provide a high-level summary of how the reprojection code works. What are the different reprojection methods and why do we prefer e.g., "block" interpolation in some places and NN in others?

seaice_ecdr/regrid_25to12.py
trey-stafford, line 1: We should add some tests to validate the logic of this code.

trey-stafford (commented):
This code is working, but there are some changes that need to be made before we merge.

The biggest stand-out items to me are:

* Refactoring how TB channel names get re-mapped (this should be done before we merge w/ `main`. We could probably merge this into `add_f17_support` though).
* Adding tests to ensure the reprojection logic works as expected

seaice_ecdr/initial_daily_ecdr.py
trey-stafford: Added a trello card: https://trello.com/c/QxOgp6pe/364-ecdr-refactor-passing-of-tbs-to-pmicecon-from-various-input-sources

seaice_ecdr/platforms.py
trey-stafford: Oh, I think I understand - F17 begins in 2008 but we don't start using it until 2011, correct? I think we should still remove this comment. The platform availability map should make this clear.

seaice_ecdr/ancillary.py
trey-stafford: This is because in some cases, we have a lower case f17 vs F17.

We need to remove this logic and correct the problem upstream.

PR #76, opened 2024-01-19
Add CLI for validating outputs

trey-stafford:
To create daily and monthly log files for the N and S hemisphere over a date range:

```
$ ./scripts/cli.sh validate-outputs --start-date 2022-01-01 --end-date 2022-01-30
```


seaice_ecdr/util.py
```diff
@@ -98,3 +99,9 @@ def sat_from_filename(filename: str) -> SUPPORTED_SAT:
     sat = cast(SUPPORTED_SAT, sat)
 
     return sat
+
+
+def date_range(*, start_date: dt.date, end_date: dt.date) -> Iterator[dt.date]:
```
trey-stafford, line 104: I realize this is already defined in `pm_icecon` but for a util like this I think I prefer having a local copy. I don't want to be dependent on `pm_icecon` for anything but the ice concentration algorithms.

seaice_ecdr/validation.py
```diff
@@ -0,0 +1,505 @@
...
+
+        flag_value = flag_value_for_meaning(
+            var=ds.surface_type_mask,
+            meaning=flag,
+        )
+        num_flag_pixels = int((ds.surface_type_mask == flag_value).sum())
+        surf_value_counts[flag] = num_flag_pixels
+
+    # Number of oceanmask (invalid ice mask) pixels
+    # TODO: inconsistency in bitmask meaning between daily & monthly.
```
trey-stafford, line 202: Does this need to be fixed, or is there a reason for the difference?

seaice_ecdr/validation.py
```diff
@@ -0,0 +1,505 @@
...
+    )
+    invalid_ice_mask = (qa_var & invalid_ice_bitmask_value) > 0
+    num_oceanmask_pixels = int(invalid_ice_mask.sum())
+
+    # Ice-free pixels (conc == 0)
+    num_ice_free_pixels = int((seaice_conc_var == 0).sum())
+
+    # TODO: we don't have missing pixels. Remove? Some other measure?
+    # Leave hard-codded to 0?
+    num_missing_pixels = 0
```
trey-stafford, line 220: Should we report the # of missing pixels? Remove from log output all together? Look at e.g., which pixels had to be spatially or temporally interpolated instead?

seaice_ecdr/validation.py
```diff
@@ -0,0 +1,562 @@
...
+            var=ds.surface_type_mask,
+            meaning=flag,
+        )
+        num_flag_pixels = int((ds.surface_type_mask == flag_value).sum())
+        surf_value_counts[flag] = num_flag_pixels
+
+    # Number of oceanmask (invalid ice mask) pixels
+    # TODO: inconsistency in bitmask meaning between daily & monthly.
+    _bitmask_meaning = (
+        "valid_ice_mask_applied"
```
trey-stafford, line 253: We want to make this consistent, and it should be "invalid_ice_mask_applied".

PR #79, opened 2024-01-23
Allow raw nt conc over 100

trey-stafford:
* Replace `valid_range` with `valid_min`
* Stop setting values > 200 to `np.nan`. This is old code to remove flags that are not present in the NT field anyway.


trey-stafford:
We probably want to raise an error if values >200 occur. Something for us to investigate further.

trey-stafford:
Values > 200 _do_ occur. We may need a different solution.

E.g., NH 2022-02-24.

```
ipdb> nt_conc[nt_conc > 200]
array([227.91939591, 201.75049427, 201.12848806])
```

trey-stafford:
Updated code to only set a `valid_min`. Now we log a warning if values > 200 are found.

PR #82, opened 2024-01-25
Clamp bt and CDR concentrations to 10-100.

trey-stafford:
And simplify masking of cdr conc values.

CDR SIC values are clamped to 10-100%. BT SIC values must be >= 0, but can be > 100.


seaice_ecdr/initial_daily_ecdr.py
```diff
@@ -675,9 +678,10 @@ def compute_initial_daily_ecdr_dataset(
             )
             logger.info("Updated spatial_interpolation with pole hole value")
 
-    # Apply land flag value and clamp max conc to 100.
+    # Mask out land and clamp conc to between 10-100%.
```
trey-stafford, line 681: Make a TODO to double-check this.

PR #85, opened 2024-01-26
Cleanup default flag value

trey-stafford:
We do not need to keep track of a `missing_value`. This is only used by the bootstrap algorithm and gets overwritten with `np.nan`.

Also, a few other items I noticed along the way.

### DIFF
--- a/seaice_ecdr/initial_daily_ecdr.py
+++ b/seaice_ecdr/initial_daily_ecdr.py
@@ -665,25 +665,22 @@ def compute_initial_daily_ecdr_dataset(
     cdr_conc[cdr_conc > 100] = 100
 
     # Add the BT raw field to the dataset
-    if bt_conc is not None:
-        bt_conc = bt_conc / 100.0  # re-set range from 0 to 1
-        ecdr_ide_ds["raw_bt_seaice_conc"] = (
-            ("time", "y", "x"),
-            np.expand_dims(bt_conc, axis=0),
-            {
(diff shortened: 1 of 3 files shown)


seaice_ecdr/initial_daily_ecdr.py
```diff
@@ -91,12 +86,16 @@ def cdr_bootstrap(
...
-        missing_flag_value=missing_flag_value,
+        # Note: the missing value of 255 ends up getting set to `nan` below.
+        missing_flag_value=255,
     )
 
     # Se any bootstrap concentrations below 10% to 0.
     bt_conc[bt_conc < 10] = 0
 
+    # Remove bt_conc flags (e.g., missing)
+    bt_conc[bt_conc > 200] = np.nan
```
trey-stafford, line 97: Moved this step from a different part of the code so it's more clear.

seaice_ecdr/initial_daily_ecdr.py
```diff
@@ -672,27 +664,22 @@ def compute_initial_daily_ecdr_dataset(
     cdr_conc[cdr_conc > 100] = 100
 
     # Add the BT raw field to the dataset
-    if bt_conc is not None:
```
trey-stafford, line 675: `bt_conc` and `nt_conc` are never `None`, so I removed these checks.

sc0tts (approved):
Cleaner code.  I like it!

PR #86, opened 2024-01-27
Refactor remapping tbs

trey-stafford:
Refactor how Tbs are remapped for the ECDR and passing on to `pm_icecon`.

Prior to these changes, we assumed input tbs had channel names consistent with AMSR, which were then mapped onto Tb channels expected by `pm_icecon` and the melt algorithm.

Now we expect just one consistent set of Tb channel names. It is the responsibility of the `seaice_ecdr.tb_data` module to provide data for those channels.

### DIFF
--- a/seaice_ecdr/tb_data.py
+++ b/seaice_ecdr/tb_data.py
@@ -4,6 +4,7 @@ from pathlib import Path
 from typing import Final, get_args
 
 import numpy as np
+import numpy.typing as npt
 import xarray as xr
 from loguru import logger
 from pm_tb_data._types import Hemisphere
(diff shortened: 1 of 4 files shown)


seaice_ecdr/tb_data.py
```diff
@@ -254,7 +274,8 @@ def _get_nsidc_0001_tbs(
         )
 
     ecdr_tb_data = EcdrTbData(
-        tbs=xr_tbs,
+        # TODO: does 0001 need a different mapping?
```
trey-stafford, line 277: Possible TODO.

sc0tts (approved):
Warms my heart to have the old, familiar ssmi/ssmis channel names back!

PR #90, opened 2024-01-31
Transform most Tbs to match F13

trey-stafford:
Associated PR: https://github.com/nsidc/pm_icecon/pull/41


seaice_ecdr/initial_daily_ecdr.py
```diff
@@ -68,6 +71,24 @@ def cdr_bootstrap(
     itp_37h = bt_coefs["bt_itp_h37"]
     itp_19v = bt_coefs["bt_itp_v19"]
 
+    # Transform TBs for the bootstrap calculation
```
trey-stafford, line 74: Currently, we do this twice. Once here, and then again for the bootstrap weather filter. Ideally we only do this operation once. Can come back and cleanup later!

PR #96, opened 2024-02-01
Initial ecdr runs

sc0tts:
This PR adds the ability to run days with observations from F08, F11, and F13 and also includes some minor adjustments to the code, including use of a slightly more aggressive set of NT2 land spillover options.

### DIFF
--- a/seaice_ecdr/initial_daily_ecdr.py
+++ b/seaice_ecdr/initial_daily_ecdr.py
@@ -357,6 +357,9 @@ def compute_initial_daily_ecdr_dataset(
         )
         spatint_bitmask_arr[land_mask.data] = 0
 
+        # Drop the un-spatially interpolated TB field to save space and compute
+        ecdr_ide_ds = ecdr_ide_ds.drop_vars(tb_varname)
+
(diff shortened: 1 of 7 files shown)


seaice_ecdr/initial_daily_ecdr.py
```diff
@@ -283,6 +286,8 @@ def compute_initial_daily_ecdr_dataset(
     hemisphere: Hemisphere,
     tb_data: EcdrTbData,
     fill_the_pole_hole: bool = False,
+    # SS: Does this need to have tb_resolution?
+    # tb_resolution: ECDR_SUPPORTED_RESOLUTIONS | None = None,
```
trey-stafford, line 290: This can eb removed.

seaice_ecdr/temporal_composite_daily.py
```diff
@@ -355,17 +355,21 @@ def read_or_create_and_read_idecdr_ds(
     )
     if overwrite_ide or not ide_filepath.is_file():
         excluded_idecdr_fields = [
```
trey-stafford, line 357: We probably want to make this an inclusion list. We know exactly what we want in the final output files, and everything else saved to the intermediate files is useful for debugging/dev work.

PR #97, opened 2024-02-01
Smmr weather filter

trey-stafford:
And a few other fixes that were necessary for tests to pass.

Note that I plan to look at the mypy version mismatch issue again later...hopefully we can finally get that sorted out.

### DIFF
--- a/seaice_ecdr/initial_daily_ecdr.py
+++ b/seaice_ecdr/initial_daily_ecdr.py
@@ -284,6 +284,16 @@ def get_nasateam_weather_mask(
     ecdr_ide_ds: xr.Dataset,
     nt_coefs: NtCoefs,
 ) -> npt.NDArray[np.bool_]:
+    """Return the "nasateam" weather mask.
+
+    TODO: the nasateam weather mask is the combination of two seaparate masks:
+    one generated from the ratio between v37 and v19 and the other generated
+    from a ratio between the v22 and v19 channels. When data are available to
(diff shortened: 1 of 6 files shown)


seaice_ecdr/initial_daily_ecdr.py
```diff
@@ -278,6 +279,35 @@ def _setup_ecdr_ds(
     return ecdr_ide_ds
 
 
+def get_nasateam_weather_mask(
```
trey-stafford, line 282: Extracted this logic from `compute_initial_daily_ecdr_dataset` and added in handling for SMMR.

seaice_ecdr/initial_daily_ecdr.py
```diff
@@ -278,6 +279,35 @@ def _setup_ecdr_ds(
...
+    if ecdr_ide_ds.platform == "n07":
+        return nt_3719_weather_mask
+
+    nt_gr_2219 = nt.compute_ratio(
+        ecdr_ide_ds["v22_day_si"].data[0, :, :],
+        ecdr_ide_ds["v19_day_si"].data[0, :, :],
+    )
+    nt_2219_weather_mask = nt_gr_2219 > nt_coefs["nt_gradient_thresholds"]["2219"]
+
+    weather_mask = nt_3719_weather_mask | nt_2219_weather_mask
```
trey-stafford, line 306: For all other platforms, we compute the ratio/mask for 2219. The weather mask is where either 3719 or 2219 are masked.

PR #101, opened 2024-02-02
Support SH runs

sc0tts:
There were some issues with running the southern hemisphere, eg that the pole_mask field does not occur there and therefore doesn't need to be regridded from 25km to 12.5km.

Also, because the complete_daily file generation process uses the prior-day's melt_onset field, the complete_daily files cannot be created in a multiprocessor batch...because the order matters.  (At least from March - Sep which is the melt season).  

So I commented out that part of the multiprocess loop.  This could be replaced with, say, a year-by-year set of processes?

### DIFF
--- a/seaice_ecdr/daily_aggregate.py
+++ b/seaice_ecdr/daily_aggregate.py
@@ -159,7 +159,7 @@ def make_daily_aggregate_netcdf_for_year(
     "--year",
     required=True,
     type=int,
-    help="Year for which to create the monthly file.",
+    help="Year for which to create the daily-aggregate file.",
 )
 @click.option(
     "-h",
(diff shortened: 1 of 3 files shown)


seaice_ecdr/daily_aggregate.py
```diff
@@ -159,7 +159,7 @@ def make_daily_aggregate_netcdf_for_year(
     "--year",
     required=True,
     type=int,
-    help="Year for which to create the monthly file.",
+    help="Year for which to create the daily-aggregate file.",
```
trey-stafford, line 162: Oof, good catch!

trey-stafford (approved):

PR #112, opened 2024-02-13
Create checksums files for new complete data

trey-stafford:
Adds code to generate checksum files when producing complete/final nc files. For our beta release, we just ran the `checksums` module as a script and generated the checksum files all at once, after we had generated the data. Operationally, we'll generate checksum files as new data are produced.


sc0tts (approved):
This looks good to me.  Will be nice to have the checksums generated at runtime.

PR #114, opened 2024-02-13
Fix multiprocess daily

trey-stafford:


seaice_ecdr/multiprocess_daily.py
```diff
@@ -15,6 +16,35 @@
 from seaice_ecdr.util import date_range
 
 
+def get_dates_by_year(dates: list[dt.date]) -> list[list[dt.date]]:
+    years = np.unique([date.year for date in dates])
+    dates_by_year = {}
+    for year in years:
+        dates_by_year[year] = [date for date in dates if date.year == year]
+
+    dates_by_year_list = list(dates_by_year.values())
```
sc0tts, line 25: I think we should sort the list of dates in each year. 

If we run, say, June 1st as our first date.  It will compute all of the complete daily fields back to the start of the melt season.  Then, all of the May, April, etc dates will already have been created, and we will be needlessly checking-if-they-exist-and-then-not-creating-them or (somewhat wastefully) creating them again if overwrite is set to "Do It"

sc0tts (changes requested):
The only thing I suggest is sorting the dates in each year so that they process in calendar order.

seaice_ecdr/multiprocess_daily.py
trey-stafford: dd689fc

PR #115, opened 2024-02-13
Address more TODOs and refactor tb_data module

trey-stafford:
Address some additional TODOs (e.g., removing some commented out/OBE code).

Also implements a refactoring of the tb_data module. DRYed out some code around creating null fields when data are missing and simplified how channels from input sources are mapped to ECDR/`pm_icecon` expected channel names. I added some tests around fetching and mapping tbs in support of this work as well.


sc0tts (approved):
I like what I see here.  Nice work finding another large block of copied/replaced code that was wrapped in a ''' block!

PR #117, opened 2024-02-16
Lance nrt temporal interpolation

trey-stafford:


seaice_ecdr/nrt.py
```diff
@@ -138,50 +258,19 @@ def download_latest_nrt_data(*, output_dir: Path, overwrite: bool) -> None:
...
-        platform=platform,
         ecdr_data_dir=ecdr_data_dir,
-        resolution="12.5",
-    )
-
-    write_ide_netcdf(
-        ide_ds=nrt_initial_ecdr_ds,
-        output_filepath=output_path,
+        # TODO: CLI option.
+        overwrite=True,
```
sc0tts, line 273: We probably want the overwrite default to be False.

In general, we won't want to reprocess either the LANCE-derived NRT data products or the final AU_SI12-derived products.  So we will want to be explicit about overwriting previously generated data files.

sc0tts (approved):
I made a couple of suggestions.  One about the use of LANCE_RESOLUTION as a variable/constant name.  The other about the default setting of overwrite.

Other than those suggestions -- and I can be talked out of "demanding" either or both -- this looks good to me.

seaice_ecdr/nrt.py
```diff
@@ -49,16 +73,16 @@ def compute_nrt_initial_daily_ecdr_dataset(
         ),
         xr_tbs=xr_tbs,
         hemisphere=hemisphere,
-        resolution=tb_resolution,
+        resolution=LANCE_RESOLUTION,
```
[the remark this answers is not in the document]
trey-stafford: yeah, I set this up for the sake of simplicity as I was refactoring things. We don't currently need to support anything other than 12.5km. We've "built-in" support for other resolutions in places, but we'd need to do more work to actually support anything other than 12.5km.

seaice_ecdr/nrt.py
trey-stafford: Agreed, I'll add this as a CLI option. Its useful in dev, at least. I'll probably make this change on the other branch, which adds support for the complete daily files, and resolve this comment here. (see https://github.com/nsidc/seaice_ecdr/pull/118/)

sc0tts (approved):
This looks good.  Especially because the default number of days to look back-in-the-past is changed from 4 to 5 in PR #118.

PR #118, opened 2024-02-17
Lance complete daily

trey-stafford:
This PR completes the work started in #117 . This PR focuses on implementing the complete daily processing for NRT data, but also touches some elements of the temporal interpolation code as well.

To support NRT processing, some refactoring was required. The complete daily processing included steps for standard (non-NRT) processing that assumed the standard input data sources. This PR begins the work of extracting the logic necessary to generate dependent data (e.g., previous days' worth of data during the melt season) to a higher level, so that e.g., the logic to update the melt field for a given day doesn't need to be recursive. 

I also began  renaming functions that are specific to standard data inputs/processing to indicate if they are "standard" vs NRT, so that it's a little clearer which functions have built-in assumptions about how to create new data if it doesn't already exist.

There's more work that could be done here to simplify the logic and separate concerns (e.g., I'd like to split out logic that requires touching data on disk from performing operations on the resultant data more thoroughly). The work I did to add "standard" to function names in the complete daily module could also be applied to the temporal interpolation module. However, this PR still represents an improvement and the code works as expected  in my testing. Standard processing should continue to work as expected.

Still TODO: add attribute to complete daily conc fields indicating the number of missing pixels.

### DIFF
--- a/seaice_ecdr/initial_daily_ecdr.py
+++ b/seaice_ecdr/initial_daily_ecdr.py
@@ -998,7 +998,7 @@ def make_idecdr_netcdf(
     hemisphere: Hemisphere,
     resolution: ECDR_SUPPORTED_RESOLUTIONS,
     ecdr_data_dir: Path,
-    excluded_fields: Iterable[str] = [],
+    excluded_fields,
 ) -> None:
     logger.info(f"Creating idecdr for {date=}, {hemisphere=}, {resolution=}")
     ide_ds = initial_daily_ecdr_dataset(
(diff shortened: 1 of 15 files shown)


seaice_ecdr/initial_daily_ecdr.py
```diff
@@ -998,7 +998,7 @@ def make_idecdr_netcdf(
     hemisphere: Hemisphere,
     resolution: ECDR_SUPPORTED_RESOLUTIONS,
     ecdr_data_dir: Path,
-    excluded_fields: Iterable[str] = [],
```
trey-stafford, line 1001: We have a card to address this issue, but I ran into it again during this work. I forgot to set this during nrt processing, which resulted in unexpected fields. I removed the default arg here so that I'm at least forced to remember this consideration.

tasks/test.py
trey-stafford, line 1: I was able to increase our coverage limits, so I did!

sc0tts (approved):
I looked through the comments in the PR and the diff's of the files that changed.

This all looks good to me.

PR #120, opened 2024-02-19
Initrun v05r01

sc0tts:
Found a few "gotchas" when running the entire v05r01 set of data.  Some of these gotchas are simply identified in this PR, but are *not fixed* in code.

- Picking a mid-summer Southern Hemisphere date causes files to be created since the beginning of the melt-onset time period, but there is not melt_onset field in the SH, so all of those files are not necessary.  (issue identified, but not addressed)
- Southern Hemisphere complete_daily Datasets were being processed without the finalize_...() routine being called.  (Fixed, but should probably be refactored)
- The multiprocess-daily cli option fails if the 5-day data period extends before the start of the satellite record, which is Oct 25, 1978.  (identified, but not fixed)

One last item:  This code will use all available bytes on the storage system and then crash.  Is there a better behavior?


sc0tts:
Another thing I remembered:  I think I saw files get re-generated even though they already existed.  We should verify that the "overwrite" flag defaults to False.

sc0tts:
Another ease-of-use suggestion:  The daily-aggregate functionality is both time-consuming and parallelizable.  Perhaps add a CLI option to take advantage of multiple cores to shorten the clock-time for production?

seaice_ecdr/cli/entrypoint.py
```diff
@@ -6,6 +6,13 @@
 from seaice_ecdr.initial_daily_ecdr import cli as ecdr_cli
 from seaice_ecdr.monthly import cli as monthly_cli
 from seaice_ecdr.monthly_aggregate import cli as monthly_aggregate_cli
+
+# TODO: The multiprocess daily invocation causes an error if the
+#       temporal_composite_daily interpolation attempts to access
+#       a day prior to the start of SMMR (10/25/1978)
+# TODO: I think the overwrite flag might not be defaulting to False,
```
trey-stafford, line 13: I looked into this. I fixed one small issue with the overwrite flag for the NRT CLI, but otherwise everything was set as expected (overwrite is `False` by default).

Note that the monthly and aggregate CLI will always overwrite existing data.

seaice_ecdr/platforms.py
```diff
@@ -223,6 +223,12 @@ def get_platform_by_date(
     date: dt.date,
 ) -> SUPPORTED_SAT:
     """Return the platform for this date."""
+    # TODO: There are a couple of date ranges for which we do not want
```
trey-stafford, line 226: I created a separate card for this TODO. I don't think the logic for this belongs to this function, so I will remove the TODO from here.

PR #121, opened 2024-02-19
Initrun v05r01 trey

trey-stafford:
Resolve issues related to #120


seaice_ecdr/cli/entrypoint.py
```diff
@@ -6,13 +6,6 @@
 from seaice_ecdr.initial_daily_ecdr import cli as ecdr_cli
 from seaice_ecdr.monthly import cli as monthly_cli
 from seaice_ecdr.monthly_aggregate import cli as monthly_aggregate_cli
-
-# TODO: The multiprocess daily invocation causes an error if the
-#       temporal_composite_daily interpolation attempts to access
-#       a day prior to the start of SMMR (10/25/1978)
-# TODO: I think the overwrite flag might not be defaulting to False,
```
trey-stafford, line 13: Confirmed we have this set properly everywhere. I did fix one issue related to this in the `nrt` module, but nothing that touched standard processing. If you see this again, it's a more subtle bug than just a flag being set to `True` someplace accidentally.

seaice_ecdr/cli/entrypoint.py
```diff
@@ -6,13 +6,6 @@
 from seaice_ecdr.initial_daily_ecdr import cli as ecdr_cli
 from seaice_ecdr.monthly import cli as monthly_cli
 from seaice_ecdr.monthly_aggregate import cli as monthly_aggregate_cli
-
-# TODO: The multiprocess daily invocation causes an error if the
```
trey-stafford, line 10: Resolved in the multiprocessing module.

seaice_ecdr/complete_daily_ecdr.py
```diff
@@ -222,8 +222,11 @@ def create_melt_onset_field(
     hemisphere: Hemisphere,
     resolution: ECDR_SUPPORTED_RESOLUTIONS,
     ecdr_data_dir: Path,
-) -> npt.NDArray[np.uint8] | None:
-    """Return a uint8 melt onset field (NH only).
+) -> npt.NDArray[np.uint8]:
```
trey-stafford, line 225: I felt this was cleaner than returning `None` if the hemisphere is "north". Instead, this function raises an error if "north" is passed to it. 

This makes it easier for me to recognize at a glance that there aren't any other cases that might generate a `None` result that I need to handle in the code. In this case, it's just a matter of doing this for the NH only - a check that can be performed at a different level of the code.

seaice_ecdr/complete_daily_ecdr.py
```diff
@@ -306,22 +311,16 @@ def create_melt_onset_field(
     return updated_melt_onset
 
 
-def complete_daily_ecdr_ds(
+def _add_melt_onset_for_nh(
```
trey-stafford, line 314: My goal with this and the `_add_surfacetype_da` function (below) was to make the `complete_daily_ecdr_ds` function easier to read and understand. That'll be more evident below.

seaice_ecdr/complete_daily_ecdr.py
```diff
@@ -350,58 +401,59 @@ def complete_daily_ecdr_ds(
...
-                "Value of 255 means no melt detected yet or the date is"
-                " outside the melt season.  Other values indicate the day"
-                " of year when melt was first detected at this location."
-            ),
-        },
-        {
-            "zlib": True,
-        },
-    )
+def complete_daily_ecdr_ds(
```
trey-stafford, line 414: The content of this function is now greatly reduced and it's easier to see what steps are being taken and why (we initialize the complete dataset w/ the temporally interpolated input dataset, add a surfacetype variable, add melt onset info for the NH only, and then finalization of the dataset.

seaice_ecdr/daily_aggregate.py
```diff
@@ -31,13 +32,7 @@ def _get_daily_complete_filepaths_for_year(
...
 ) -> list[Path]:
     data_list = []
-    # This is the beginning of the satellite record, so we can skip anything
-    # before this date when 1978 is given.
-    # TODO: pull this start date from the platforms config.
-    if year == 1978:
-        start_date = dt.date(1978, 10, 25)
-    else:
-        start_date = dt.date(year, 1, 1)
+    start_date = max(dt.date(year, 1, 1), get_first_platform_start_date())
```
trey-stafford, line 35: There were a few places where we hard-coded 1978-10-25 as the earliest start date. This ensures that the earliest start date is derived from the platform start date config (which we have support for overriding!).

seaice_ecdr/multiprocess_daily.py
```diff
@@ -104,7 +105,8 @@ def cli(
     # craete a range of dates that includes the interpolation range. This
     # ensures that data expected for temporal interpolation of the requested
     # dates has the data it needs.
-    initial_start_date = start_date - dt.timedelta(days=5)
+    earliest_date = get_first_platform_start_date()
+    initial_start_date = max(earliest_date, start_date - dt.timedelta(days=5))
```
trey-stafford, line 109: Here's where the multiprocessing bug gets fixed. I had this setup to always try to generate initial daily data for 5 days before the given start date. That's problematic at the beginning of the timeseries! The new `get_first_platform_start_date` provides us with the answer.

seaice_ecdr/nrt.py
```diff
@@ -319,7 +319,7 @@ def nrt_ecdr_for_day(
             hemisphere=hemisphere,
             date=date,
             ecdr_data_dir=ecdr_data_dir,
-            overwrite=True,
+            overwrite=overwrite,
```
trey-stafford, line 322: This is the only place I could find where an overwrite flag wasn't being passed properly.

seaice_ecdr/platforms.py
```diff
@@ -2,6 +2,14 @@
 
 Routines for dealing with the platform (satellite) and sensors
 for CDRv5
+
+
+TODO: There are a couple of date ranges for which we do not want
```
trey-stafford, line 7: Moved this TODO here (and added a trello card to address separately).

PR #122, opened 2024-02-20
Move all logging-related config to `seaice_ecdr` init file

trey-stafford:
This places logging config in `seaice_ecdr/__init__.py`. 

We have had this same config spread across multiple modules. Now it just exists in one place, resulting in a consistent logging output across the entire codebase.

Currently, this defaults the log-level to WARNING (meaning that info-level logs are no longer printed). This seems reasonable for operational use. Ideally this gets switched to "DEBUG" or something more verbose in dev environments by default. For now, we can manually adjust this as needed.

### DIFF
--- a/seaice_ecdr/__init__.py
+++ b/seaice_ecdr/__init__.py
@@ -1,1 +1,11 @@
-"""Blank __init__.py."""
+import sys
+
+from loguru import logger
+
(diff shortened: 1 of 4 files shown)


sc0tts (approved):
So much cleaner!

PR #123, opened 2024-02-20
cause multiprocess to handle overwrite flag

sc0tts:
The multiprocess cli runs pools of initial_daily, temporal_interp, and complete_daily sequentially.  The means that it will end up generating all underlying files.

Because of this, the existence of a complete_daily file does not prevent the multiprocess code from generating the initial_daily and temporal_interp files.

This PR causes initial_daily and temporal_interp to respect the overwrite flag passed to the multiprocess cli.


trey-stafford (approved):
Looks good to me.

PR #124, opened 2024-02-20
Check for missing data pixels

trey-stafford:
* Validation code now checks explicitly for missing pixels (`nan` values in the ocean mask)
* Sea ice CDR concentration variables in the daily and monthly files have a `number_of_missing_pixels` attribute with the count of missing pixels (`nan` values in the ocean mask) .

### DIFF
--- a/seaice_ecdr/daily_aggregate.py
+++ b/seaice_ecdr/daily_aggregate.py
@@ -92,6 +92,15 @@ def _update_ncrcat_daily_ds(
     ds["latitude"] = surf_geo_ds.latitude
     ds["longitude"] = surf_geo_ds.longitude
 
+    # Remove the "number of missing pixels" attr from the daily aggregate conc
+    # variable.
+    ds["cdr_seaice_conc"].attrs = {
+        k: v
+        for k, v in ds["cdr_seaice_conc"].attrs.items()
+        if k != "number_of_missing_pixels"
+    }
+    # ds["cdr_seaice_conc]
+
     # setup global attrs
     # Set global attributes
     daily_aggregate_ds_global_attrs = get_global_attrs(
--- a/seaice_ecdr/daily_aggregate.py
+++ b/seaice_ecdr/daily_aggregate.py
@@ -99,7 +99,6 @@ def _update_ncrcat_daily_ds(
         for k, v in ds["cdr_seaice_conc"].attrs.items()
         if k != "number_of_missing_pixels"
     }
-    # ds["cdr_seaice_conc]
 
     # setup global attrs
     # Set global attributes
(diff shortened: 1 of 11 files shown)


seaice_ecdr/daily_aggregate.py
```diff
@@ -92,6 +92,15 @@ def _update_ncrcat_daily_ds(
...
     ds["longitude"] = surf_geo_ds.longitude
 
+    # Remove the "number of missing pixels" attr from the daily aggregate conc
+    # variable.
+    ds["cdr_seaice_conc"].attrs = {
+        k: v
+        for k, v in ds["cdr_seaice_conc"].attrs.items()
+        if k != "number_of_missing_pixels"
+    }
+    # ds["cdr_seaice_conc]
```
sc0tts, line 102: Is this commented-out line delete-able?

seaice_ecdr/monthly.py
```diff
@@ -335,22 +339,32 @@ def calc_qa_of_cdr_seaice_conc_monthly(
...
+    num_missing_conc_pixels = get_num_missing_pixels(
+        seaice_conc_var=conc_monthly,
+        hemisphere=hemisphere,
+        resolution=resolution,
+    )
+
+    conc_monthly.name = "cdr_seaice_conc_monthly"
     conc_monthly = conc_monthly.assign_attrs(
-        long_name=long_name,
+        long_name="NOAA/NSIDC Climate Data Record of Passive Microwave Monthly Northern Hemisphere Sea Ice Concentration",
```
sc0tts, line 360: Question: is it better to have this type of thing in the code here, or to have it in a configuration file somewhere?

seaice_ecdr/daily_aggregate.py
trey-stafford: ooh, good catch! I'll clean that up.

trey-stafford: 21fe450

seaice_ecdr/monthly.py
trey-stafford: It's a good question. As an initial implementation throughout this code  I've tried to consolidate constants like this to places like e.g., the `nc_attrs` module. In this case, the monthly code just hard-codes this value where the variable gets created.

I think an improvement on this would be to extract information about variable attrs  and nc structure for various output files to some kind of data model (e.g. using pydantic or dataclasses) like we've discussed. Defaults for things like `long_name` could be set this way, but also be easy to find and changeable. I think this sort of item (not likely to be changed very often), can safely live in code _as configuration_. Supposing the models are well documented, even someone who has little experience with code could make tweaks to that.

Extracting this to a configuration file outside of the code would be, imo, more of a needs-based consideration. Do we need to be able to change this value along with others by swapping in alternative configurations in different environments? For example, if we had planned to produce two completely separate CDRs using completely different input sources and publish those using a separate VM, it would be nice to have this sort of thing extracted out to an external config file that can be different in each instance.

PR #128, opened 2024-02-22
Improve error handling for primary ECDR outputs

trey-stafford:
This PR improves how errors and log messages are handled throughout the code.

* Use `logger.debug` more extensively, replacing the nearly universal use of `logger.info`. DEBUG level log statements will appear in dev, but not in other environments (set `ENVIRONMENT=dev` to achieve this verbose logging).
* Errors are raised if errors are encountered, resulting in non-zero exit codes. I've tried to replicate the original behavior (where it mattered) of e.g., for the complete daily CLI, letting failures "go" until after an entire range of dates is completed. Instead of just logging these errors, an error gets raised after all dates have been attempted, if an error was encountered for any of the dates.

### DIFF
--- a/seaice_ecdr/__init__.py
+++ b/seaice_ecdr/__init__.py
@@ -9,3 +9,4 @@ except ValueError:
     logger.debug(f"Started logging in {__name__}")
 
 logger.add(sys.stderr, level="WARNING")
+logger.add(sys.stdout, level="INFO")
--- a/seaice_ecdr/__init__.py
+++ b/seaice_ecdr/__init__.py
@@ -3,12 +3,11 @@ import sys
 
 from loguru import logger
 
-# Set the default minimum log notification to Warning
-try:
-    logger.remove(0)  # Removes previous logger info
-except ValueError:
-    logger.debug(f"Started logging in {__name__}")
-
 env = os.environ.get("ENVIRONMENT")
 _default_log_level = "DEBUG" if env == "dev" else "INFO"
-logger.add(sys.stdout, level=_default_log_level)
+
+logger.configure(
+    handlers=[
+        dict(sink=sys.stdout, level=_default_log_level),
+    ]
+)
--- a/seaice_ecdr/__init__.py
+++ b/seaice_ecdr/__init__.py
@@ -8,6 +8,6 @@ _default_log_level = "DEBUG" if env == "dev" else "INFO"
 
 logger.configure(
     handlers=[
-        dict(sink=sys.stdout, level=_default_log_level),
+        dict(sink=sys.stderr, level=_default_log_level),
     ]
 )
--- a/seaice_ecdr/__init__.py
+++ b/seaice_ecdr/__init__.py
@@ -1,3 +1,4 @@
+import os
 import sys
 
 from loguru import logger
@@ -8,5 +9,6 @@ try:
 except ValueError:
     logger.debug(f"Started logging in {__name__}")
 
-logger.add(sys.stderr, level="WARNING")
-logger.add(sys.stdout, level="INFO")
+env = os.environ.get("ENVIRONMENT")
+_default_log_level = "DEBUG" if env == "dev" else "INFO"
+logger.add(sys.stdout, level=_default_log_level)
--- a/seaice_ecdr/complete_daily_ecdr.py
+++ b/seaice_ecdr/complete_daily_ecdr.py
@@ -734,7 +734,6 @@ def cli(
     error_dates = []
     for date in date_range(start_date=date, end_date=end_date):
         try:
-            1 / 0
             create_standard_cdecdr_for_date(
                 hemisphere=hemisphere,
                 date=date,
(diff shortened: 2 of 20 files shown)


seaice_ecdr/__init__.py
```diff
@@ -1,11 +1,13 @@
...
-    logger.remove(0)  # Removes previous logger info
-except ValueError:
-    logger.debug(f"Started logging in {__name__}")
+env = os.environ.get("ENVIRONMENT")
+_default_log_level = "DEBUG" if env == "dev" else "INFO"
 
-logger.add(sys.stderr, level="WARNING")
+logger.configure(
+    handlers=[
+        dict(sink=sys.stdout, level=_default_log_level),
```
trey-stafford, line 11: We had been logging to `stderr`. I set to `stdout`, because this is what I expected. Ideally errors & warnings get sent to `stderr` and everything else goes to `stdout`, but it wasn't clear how to achieve that without some customization.

seaice_ecdr/__init__.py
```diff
@@ -1,11 +1,13 @@
...
+    ]
+)
```
trey-stafford, line 13: TODO: One thing I still want to add to this configuration is logging to a file. We write out log error files for individual outputs, but that's only the exception, and not any of the surrounding logged information.

seaice_ecdr/complete_daily_ecdr.py
```diff
@@ -83,7 +87,7 @@ def read_tiecdr_ds(
         resolution=resolution,
         ecdr_data_dir=ecdr_data_dir,
     )
-    logger.info(f"Reading tieCDR file from: {tie_filepath}")
+    logger.debug(f"Reading tieCDR file from: {tie_filepath}")
```
trey-stafford, line 90: We should use `debug` to log information that might be handy in dev, but that a normal user (e.g., ops) wouldn't want to see. 

Given ops is our primary user, info-level logs should contain information useful for...ops. This includes stuff like progress updates for long-running tasks and when/where files are written out to.

seaice_ecdr/complete_daily_ecdr.py
```diff
@@ -631,65 +642,39 @@ def read_or_create_and_read_standard_cdecdr_ds(
...
-    for date in date_range(start_date=start_date, end_date=end_date):
-        create_standard_cdecdr_for_date(
-            hemisphere=hemisphere,
-            date=date,
-            resolution=resolution,
-            ecdr_data_dir=ecdr_data_dir,
-            overwrite_cde=overwrite_cde,
-        )
+) -> list[dt.date]:
+    """TODO: docstring"""
```
trey-stafford, line 653: TODO!

seaice_ecdr/daily_aggregate.py
```diff
@@ -229,10 +237,31 @@ def cli(
...
+                ),
+                ecdr_data_dir=ecdr_data_dir,
+                product_type="aggregate",
+            )
+
+    if failed_years:
+        str_formatted_years = "\n".join(str(year) for year in failed_years)
+        raise RuntimeError(
+            f"Encountered {len(failed_years)} failures."
+            f" Data for the following years were not created:\n{str_formatted_years}"
```
trey-stafford, line 266: TODO: More specific error. Indicate that daily aggregate files were not created.

seaice_ecdr/initial_daily_ecdr.py
```diff
@@ -1071,45 +1068,20 @@ def create_idecdr_for_date(
...
-    """Generate the initial daily ecdr files for a range of dates."""
-    for date in date_range(start_date=start_date, end_date=end_date):
-        create_idecdr_for_date(
-            hemisphere=hemisphere,
-            date=date,
-            resolution=resolution,
-            ecdr_data_dir=ecdr_data_dir,
-            verbose_intermed_ncfile=verbose_intermed_ncfile,
-        )
+        # TODO: err logfile for intiial daily data?
```
trey-stafford, line 1075: I think we can safely remove this TODO and not worry about this, unless scott thinks it would be valuable!

seaice_ecdr/temporal_composite_daily.py
```diff
@@ -1005,10 +959,12 @@ def cli(
     if end_date is None:
         end_date = copy.copy(date)
 
+    # TODO: support 'failed' outputs from this CLI?
```
trey-stafford, line 962: TODO: depends on outcome of the above TODO about the same thing. I lean toward not creating these for intermediate files.

trey-stafford:
We should also add log success statements for normal processing.

seaice_ecdr/__init__.py
trey-stafford: Changed back to `stderr`. Scott's invocation of the CLI captures stderr logs.

seaice_ecdr/complete_daily_ecdr.py
trey-stafford: 4bb361e

PR #130, opened 2024-02-22
Persistent logging

trey-stafford:
Adds "persistent" logging.

Logs are now written to `/share/apps/G02202_V5/v05r01/logs/{YYYY-MM-DD}.log`. Where YYYY-MM-DD reflects the current date.

Logs are retained for 31 days, so if a problem occurs that's how long we'd have to look at these logs. I haven't setup long-term archival for these logs, but we could.

 _Usually_ when something goes wrong in prod w/ a CLI program like this, the Jenkins logs provide enough detail (and persistence) for us to figure out what's going on w/o needing to ever look at application logs directly anyway.

### DIFF
--- a/seaice_ecdr/__init__.py
+++ b/seaice_ecdr/__init__.py
@@ -25,6 +25,11 @@ logger.configure(
 # level?) for speed and space consideration issues.
 do_not_log = os.environ.get("DISABLE_FILE_LOGGING")
 if not do_not_log:
+    # One file per day.
     file_sink_fp = LOGS_DIR / "{time:%Y-%m-%d}.log"
-    # Retain logs for up to a month.
-    logger.add(file_sink_fp, level=DEFAULT_LOG_LEVEL, retention=31)
+    logger.add(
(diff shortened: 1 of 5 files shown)


seaice_ecdr/__init__.py
```diff
@@ -3,11 +3,28 @@
...
+# Configure the default logger, which prints to stderr.
 logger.configure(
     handlers=[
         dict(sink=sys.stderr, level=_default_log_level),
-    ]
+    ],
 )
+
+# Optionally (by default) log to a file named after the current date.
+# TODO: consider logging all messages (TRACE level)? We could retain all context for
```
trey-stafford, line 22: Tradeoff here w/ the increased verbosity. On  one hand, it might be handy to see all debug logs, but on the other, it often isn't necessary and adds clutter to the other more-important messages (e.g,. most of the time we look at this will likely be to look for exceptions!). I'll leave as a TODO for us to re-consider later.

PR #133, opened 2024-02-28
Add versioning for ECDR and update dev docs

trey-stafford:
From this point onward, we should update the CHANGELOG w/ new PRs and version the code as needed.


sc0tts (approved):
Upon cursory review -- because I don't yet "grok" the details of versioning -- this looks good.

Quick question though: would it be advisable to keep the version number the same as the version number of the data set that is being created?

PR #137, opened 2024-09-05
Use pydantic to manage platforms, and fix failing tests

trey-stafford:
Refactors how platforms are defined and handled in the code.

* Updates naming of function arg/kwargs and variables from `sat` to `platform_id`
* Defines pydantic models for platform-related configuration
* Platform start dates for the CDR are defined by a YAML configuration file. This configuration file can be overridden with an environment variable (`PLATFORM_START_DATES_CONFIG_FILEPATH`)
* Fixes tests that were broken in the destination branch.

Note that `pm_icecon` and `pm_tb_data` still utilize `sat` to refer to the platform ID, but that's OK for now.


seaice_ecdr/ancillary.py
```diff
@@ -77,7 +77,7 @@ def get_surfacetype_da(
     date: dt.date,
     hemisphere: Hemisphere,
     resolution: ECDR_SUPPORTED_RESOLUTIONS,
-    platform: SUPPORTED_SAT,
+    platform: SUPPORTED_PLATFORM_ID,
```
trey-stafford, line 80: This function could take a `Platform` instead of a platform ID. At a minimum, updating the arg name here is probably a good idea.

seaice_ecdr/ancillary.py
```diff
@@ -165,7 +165,7 @@ def nh_polehole_mask(
     polehole_bitmask = ancillary_ds.polehole_bitmask
 
     if sat is None:
-        sat = get_platform_by_date(
+        sat = PLATFORM_CONFIG.get_platform_by_date(
```
trey-stafford, line 168: Should be `platform`, not `sat`.

seaice_ecdr/ancillary.py
```diff
@@ -229,7 +229,7 @@ def get_invalid_ice_mask(
     hemisphere: Hemisphere,
     date: dt.date,
     resolution: ECDR_SUPPORTED_RESOLUTIONS,
-    platform: SUPPORTED_SAT,
+    platform: SUPPORTED_PLATFORM_ID,
```
trey-stafford, line 232: Again, change the arg to take a platform or rename the arg.

PR #140, opened 2024-09-13
Daily amsr2 prototype fields

trey-stafford:
Initial code to support creating daily files with an `amsr2_prototype` subgroup for AMSR2 fields.

There are still a fair number of TODOs here, but they can wait for if we have more time after adding support for daily aggregate files, monthly files, validation, and NRT.

For usage, see `make_25km_ecdr.sh`.

Creating the AMSR2 files relies on setting an environment variable at import-time, which means that the code cannot run as a single process. First, daily data is produced using the default platform ranges. Then, a new process is spawned (see `cli/daily.py`) with the `PLATFORM_START_DATES_CONFIG_FILEPATH` set to the prototype platform start dates config (containing just AMSR2), which produces the amsr2 fields. Finally, the files get staged for publication by combining the F17 outputs w/ AMSR2 outputs in the `prototype_amsr2` group (`publish_daily.py`).

Longer term, a complete refactor of the code to support this scheme from foundations would be ideal. We could have a single config file with `cdr_platform_start_dates` and `prototype_platform_start_dates` keys that the program would know how to parse and act accordingly without needing to spawn separate processes to access different platform configs. We don't have the time for that now, and the present approach is a byproduct of development that originally did not have this in mind (we were going to produce separate 12.5km data files for AMSR2 instead of adding it as a "prototype" group to F17 output files - just a single platform start date config).


seaice_ecdr/intermediate_daily.py
```diff
@@ -401,23 +403,17 @@ def _add_melt_onset_for_nh(
...
     #       up as the temporally-interpolation-has-occured value
     #       rather than hardcoded as '128'.
-    cde_ds_with_melt_onset["qa_of_cdr_seaice_conc"] = cde_ds_with_melt_onset[
-        "qa_of_cdr_seaice_conc"
+    cde_ds_with_melt_onset["cdr_seaice_conc_qa"] = cde_ds_with_melt_onset[
+        "cdr_seaice_conc_qa"
     ].where(
         ~is_melt_has_occurred,
-        other=np.bitwise_or(cde_ds_with_melt_onset["qa_of_cdr_seaice_conc"], 128),
+        other=np.bitwise_or(cde_ds_with_melt_onset["cdr_seaice_conc_qa"], 128),
```
sc0tts, line 416: Let's add a TODO so that this flag value is pulled from a list of flag values -- ideally the same list that is used to set the attributes in the netCDF variable -- rather than being hardcoded as 128

seaice_ecdr/multiprocess_intermediate_daily.py
```diff
@@ -29,7 +29,86 @@
...
+    ancillary_source: ANCILLARY_SOURCES,
+):
+    dates = list(date_range(start_date=start_date, end_date=end_date))
+    dates_by_year = get_dates_by_year(dates)
+
+    # craete a range of dates that includes the interpolation range. This
+    # ensures that data expected for temporal interpolation of the requested
+    # dates has the data it needs.
+    earliest_date = PLATFORM_CONFIG.get_first_platform_start_date()
+    initial_start_date = max(earliest_date, start_date - dt.timedelta(days=5))
```
sc0tts, line 49: Let's add a TODO to replace the "5" here with a config variable indicating the max days to look forward/back when running temporal interpolation

seaice_ecdr/multiprocess_intermediate_daily.py
```diff
@@ -29,7 +29,86 @@
...
+    initial_end_date = min(end_date + dt.timedelta(days=5), dt.date.today())
```
sc0tts, line 50: (Same comment as above:  replace "5" with a config var.  Or at least add a TODO to remind us to do so.

seaice_ecdr/multiprocess_intermediate_daily.py
```diff
@@ -29,7 +29,86 @@
...
+        resolution=resolution,
+        base_output_dir=base_output_dir,
+        overwrite_cde=overwrite,
+        land_spillover_alg=land_spillover_alg,
+        ancillary_source=ancillary_source,
+    )
+
+    # Use 6 cores. This seems to perform well. Using the max number available
+    # can cause issues...
+    with Pool(6) as multi_pool:
```
sc0tts, line 93: Should we make a check of the operating system to ensure that the number of cores used is ... maybe max(1, (num_cores_available - 1)) ?

seaice_ecdr/publish_daily.py
```diff
@@ -0,0 +1,368 @@
...
+        try:
+            prototype_daily_ds = read_cdecdr_ds(
+                intermediate_output_dir=intermediate_output_dir,
+                resolution=resolution,
+                date=date,
+                platform_id=PROTOTYPE_PLATFORM_ID,
+                hemisphere=hemisphere,
+                is_nrt=False,
+            )
+            cdr_var_fieldnames = [
```
sc0tts, line 199: Oh man, this list looks so NICE compared to prior versions!

seaice_ecdr/tests/unit/test_monthly.py
```diff
@@ -26,50 +27,51 @@
...
         year=2022,
         is_nrt=False,
     )
     year = 2022
     month = 3
     _fake_files_for_test_year_month_and_hemisphere = [
-        nh_complete_dir / "sic_psn12.5_20220301_am2_v05r01.nc",
-        nh_complete_dir / "sic_psn12.5_20220302_am2_v05r01.nc",
-        nh_complete_dir / "sic_psn12.5_20220303_am2_v05r01.nc",
+        nh_intermediate_dir / "sic_psn12.5_20220301_F17_v05r01.nc",
```
sc0tts, line 45: Add TODO:  want to use a version string rather than hardcoded "v05r01"

seaice_ecdr/tests/unit/test_platforms.py
```diff
@@ -195,3 +195,38 @@ def test_get_first_platform_start_date():
     )
 
     assert actual == min_date
+
+
+def test_platform_available_for_date():
+    config = PlatformConfig(
+        platforms=[
+            Platform(
+                id="ame",
```
sc0tts, line 204: I don't think we're allowing AMSR-E, are we?

sc0tts (approved):
I added some single-comments to specific locations in the code where I think we should add "TODO" statements to remind us to replace hard-coded values with configuration values.

Also, there is mention of AMSR-E at at least one place in the code.  Is that OBE now?

But these are minor and should not hold up the merging of this valuable  PR!

seaice_ecdr/intermediate_daily.py
trey-stafford: There is actually already a TODO for this, but it's a couple lines above this one. 

Added a new issue to address: https://github.com/nsidc/seaice_ecdr/issues/150

seaice_ecdr/multiprocess_intermediate_daily.py
trey-stafford: 852922b

seaice_ecdr/multiprocess_intermediate_daily.py
trey-stafford: Good suggestion. 3ffabb0

seaice_ecdr/tests/unit/test_monthly.py
trey-stafford: This is just a unit test, so it doesn't matter so much, but agreed! 6da45a3

seaice_ecdr/tests/unit/test_platforms.py
trey-stafford: Not yet (maybe not ever?) but the code exists to support it. This unit test just uses ame as an example.

PR #141, opened 2024-09-13
Match CDRv4 at 25km

sc0tts:
Minor tweaks to as-exactly-as-possible match CDRv4

Includes an update to fix how platform ID is used to determine if it is before the AMSR period.

### DIFF
--- a/seaice_ecdr/temporal_composite_daily.py
+++ b/seaice_ecdr/temporal_composite_daily.py
@@ -521,12 +521,26 @@ def get_daily_climatology_mask(
     daily_ds = None
 
     if hemisphere == "north" and resolution == "25":
+        # New mask.  Valid/invalid extends into landmask
+        # daily_ds = Dataset(
+        #    "/share/apps/G02202_V5/v05r01_ancillary/ecdr-ancillary-psn25-dailyclim.nc"
+        # )
+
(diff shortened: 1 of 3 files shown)


seaice_ecdr/temporal_composite_daily.py
```diff
@@ -525,12 +521,26 @@ def get_daily_climatology_mask(
...
+        # Mask used in v04r00.  Includes landmask and may have mismatches
+        # between landmask and ocean, especially in Southern Hemisphere
         daily_ds = Dataset(
-            "/share/apps/G02202_V5/v05r01_ancillary/ecdr-ancillary-psn25-dailyclim.nc"
+            "/share/apps/G02202_V5/v05r01_ancillary/ecdr-ancillary-psn25-smmr-invalid-ice-v04r00.nc"
         )
     elif hemisphere == "south" and resolution == "25":
+        # New mask.  Valid/invalid extends into landmask
+        # daily_ds = Dataset(
+        #    "/share/apps/G02202_V5/v05r01_ancillary/ecdr-ancillary-pss25-dailyclim.nc"
```
trey-stafford, line 537: Can we cleanup this commented-out code before merging?

trey-stafford: Scott will address this in work to "undo" the cdrv4 matching.

PR #145, opened 2024-09-15
Ecdr at 25km ssmatch cd rv4

sc0tts:
This code is meant to match the CDRv4 results as closely as possible.

There are still some differences, but for the SSMI and SSMIS sensors, this appears to be roundoff differences.

In the SMMR era, there is more mismatch.


trey-stafford:
Closing this PR because it is a duplicate of #141 , except this one targets the incorrect branch.

PR #151, opened 2024-09-18
Daily aggregate update for nc groups

trey-stafford:
Following work to implement the `prototype_amsr2` and `cdr_supplementary` groups in #140 , this PR adds support for daily aggregate files that contain those groups.


sc0tts (approved):
looks good!

PR #152, opened 2024-09-18
Rename `prototype_amsr2` -> `prototype_am2`

trey-stafford:
Use the platform ID as an identifier like we do for the variable names themselves. This is more consistent and allows the code to be more flexible, which we want for the future as new platforms are developed and put into operations.


sc0tts (approved):
Looks good!

I like the am2  label instead of AMSR2

PR #153, opened 2024-09-18
Reset ECDR Product version for initial release

trey-stafford:


seaice_ecdr/util.py
```diff
@@ -24,8 +24,8 @@ def standard_daily_filename(
 ) -> str:
     """Return standard daily NetCDF filename.
 
-    North Daily files: sic_psn12.5_{YYYYMMDD}_{platform_id}_v05r01.nc
-    South Daily files: sic_pss12.5_{YYYYMMDD}_{platform_id}_v05r01.nc
+    North Daily files: sic_psn12.5_{YYYYMMDD}_{platform_id}_{cdr_product_version}.nc
```
sc0tts, line 27: Could you check this?

Above -- as I am working through the "Files Changed" on the github PR review page -- we are using:
ECDR_PRODUCT_VERSION
  which is: all-caps,  and Ecdr

but here, we are using 
cdr_product version
  which is:  all-lowercase, and (notE)cdr

But from context, I'm thinking we want these to refer to the same constant?

trey-stafford: yeah, good catch - I'll update!

sc0tts (approved):
Ah, the sweet sound of hardcoded values leaving the codebase....

PR #155, opened 2024-09-19
Monthly update for NetCDF groups

trey-stafford:


sc0tts (approved):
Good to have the monthly variable names follow the same format as the daily :+1:

PR #157, opened 2024-09-19
Add NT2_BT land spillover

sc0tts:
Adds the option to use the NT2_BT land spillover routine
...and sets that as the default.

### DIFF
--- a/seaice_ecdr/spillover.py
+++ b/seaice_ecdr/spillover.py
@@ -23,7 +23,7 @@ from seaice_ecdr.tb_data import (
 from seaice_ecdr.util import get_ecdr_grid_shape
 
 NT_MAPS_DIR = Path("/share/apps/G02202_V5/cdr_testdata/nt_datafiles/data36/maps")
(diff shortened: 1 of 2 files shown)


seaice_ecdr/spillover.py
```diff
@@ -266,6 +266,47 @@ def land_spillover(
...
+        adj123 = get_adj123_field(
+            hemisphere=hemisphere,
+            resolution=tb_data.resolution,
+            ancillary_source=ancillary_source,
+        )
+        spillover_applied_nt2 = apply_nt2_land_spillover(
+            conc=cdr_conc,
+            adj123=adj123.data,
+            l90c=l90c.data,
+            # anchoring_siconc=50.0,
```
trey-stafford, line 287: Can we remove these commented out kwargs?

trey-stafford (approved):
One question about cleaning up some commented-out kwargs, but otherwise looks good!

seaice_ecdr/spillover.py
sc0tts: Yes.  Done.

PR #160, opened 2024-09-19
Rename flag vars to include `_flag` at the end

trey-stafford:


seaice_ecdr/publish_monthly.py
```diff
@@ -104,7 +104,6 @@ def prepare_monthly_nc_for_publication(
         if f"_{PROTOTYPE_PLATFORM_ID}_" not in fp.name
     ]
     if len(default_monthly_fps) != 1:
-        breakpoint()
```
sc0tts, line 107: Lol.  This reminds of the type of thing I submit to the repo!

sc0tts (approved):

seaice_ecdr/publish_monthly.py
trey-stafford: It happens :sweat_smile: . Ideally our linter would identify these and flag them for removal. I wonder if that's something we can enable. Might look at later.

PR #163, opened 2024-09-24
NRT Processing updates for G10016

trey-stafford:
* Adds support for  the `cdr_supplementary` NC group
* Changes the output location for NRT data to `/share/apps/G10016_V3` and updates associated dataset attrs to reflect G10016 instead of G02202
* Update `ECDR_PRODUCT_VERSION` type from string to pydantic model.

### DIFF
--- a/seaice_ecdr/platforms/config.py
+++ b/seaice_ecdr/platforms/config.py
@@ -62,7 +62,6 @@ F18_PLATFORM = Platform(
     sensor="SSMIS > Special Sensor Microwave Imager/Sounder",
     id="F18",
     date_range=DateRange(
-        # TODO: is this accurate? This value from NSIDC-0001 docs.
         first_date=dt.date(2017, 1, 1),
         last_date=None,
     ),
--- a/seaice_ecdr/platforms/config.py
+++ b/seaice_ecdr/platforms/config.py
@@ -53,6 +53,18 @@ AME_PLATFORM = Platform(
     ),
 )
 
+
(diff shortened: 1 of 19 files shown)


seaice_ecdr/platforms/config.py
```diff
@@ -53,6 +53,18 @@
...
 )
 
+
+F18_PLATFORM = Platform(
+    name="DMSP 5D-3/F18 > Defense Meteorological Satellite Program-F18",
+    sensor="SSMIS > Special Sensor Microwave Imager/Sounder",
+    id="F18",
+    date_range=DateRange(
+        # TODO: is this accurate? This value from NSIDC-0001 docs.
+        first_date=dt.date(2017, 1, 1),
```
trey-stafford, line 63: This doesn't currently get used in the code but in the future would determine what the earliest date of F18 data is, if we ever want to support it beyond use in NRT processing.

sc0tts: I checked 0001.  There are no F18 fields for 12/31/2016, and there are F18 fields for 1/1/2017.

So we can remove the "TODO" here.  The start-date of 1/1/2017 for F18 is accurate.

sc0tts (approved):
Everything here looks good for adding NRT as a processing option.  Looking forward to generating files with this code!

seaice_ecdr/platforms/config.py
trey-stafford: 01ab4e4

PR #164, opened 2024-09-26
Use f17 nrt

trey-stafford:


seaice_ecdr/config/nrt_platform_start_dates.yml
```diff
@@ -1,5 +1,4 @@
 cdr_platform_start_dates:
-  # We only use F18 for NRT.
-  - platform_id: "F18"
-    # This appears to be the first date of data in NSIDC0080.
-    start_date: "2021-11-01"
+  # We only use F17 for NRT.
+  - platform_id: "F17"
+    start_date: "2011-10-04"
```
sc0tts, line 4: I think this start date should be Jan 1, 2008.
The last day of 0001 F13 data is Dec 31, 2007.
We should probably keep the NRT and Final input platforms the same if we can.

sc0tts (approved):

seaice_ecdr/config/nrt_platform_start_dates.yml
trey-stafford: @sc0tts  can you confirm this fix is correct? 194f743

PR #167, opened 2024-09-26
Remove nrt-specific output dir code

trey-stafford:
Originally we were only going to produce daily nrt files and they were going to be a part of G02202. Now they'll ne a new version of G10016 and the directory structure for nrt files should reflect this.


trey-stafford:
Results in dirs that look like this:

```
/share/apps/G10016_V3/v03r00/dev/trst2284/complete
├── north
│   ├── checksums
│   │   └── daily
│   │       └── 2024
│   └── daily
│       └── 2024
└── south
    ├── checksums
    │   └── daily
    │       └── 2024
    └── daily
        └── 2024
```

sc0tts (approved):
Sure is simpler not to have kwarg is_nrt in so many places!

PR #168, opened 2024-09-27
Monthly NRT

trey-stafford:
Adds a CLI for monthly NRT processing.

### DIFF
--- a/doc/operation.md
+++ b/doc/operation.md
@@ -1,0 +1,108 @@
+# Sea Ice E-CDR Operations
+
+This document outlines how the `seaice_ecdr` code is leveraged in operations.
+
+
+## Command-line Interface (CLI)
+
+The `./scripts/cli.sh` script is the primary entrypoint for interacting with the
+CLI. Not all CLI subcommands are used in operations (e.g., they are used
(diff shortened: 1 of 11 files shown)


trey-stafford:
This does not update any attrs to reflect the NRT nature of this product. Need to update the title/version/summary/metadata_link attrs similar to how the daily nrt code does it.

doc/operation.md
```diff
@@ -0,0 +1,108 @@
...
+
+###  Daily processing
+
+Daily NRT processing should occur by running this command:
+
+```
+daily-nrt --last-n-days 5 --hemisphere both
+```
+
+Note that the `--overwrite` flag can be used to re-create NRT data if e.g., a
```
sc0tts, line 35: Do you want to explicitly give the command to use the overwrite flag?

eg
```
daily-nrt --last-n-days 5 --hemisphere both --overwrite
```

...in this README.md ?

sc0tts (approved):
Again, always good to see the single-use kwarg is_nrt go away.

And good to see the NRT output file directory structure mirror that of the "final" processing.

PR #171, opened 2024-11-04
Nrt for f17 and am2

trey-stafford:


trey-stafford:
This works, but I'd like to make a few more changes to the CLI for Ops' sake. I'll address those in a separate PR.

PR #173, opened 2024-11-06
do not use melt_onset QA flag mask for south

sc0tts:
This PR fixes the inclusion of a "melt_onset" bitmask in the southern hemisphere monthly files.

### DIFF
--- a/seaice_ecdr/intermediate_monthly.py
+++ b/seaice_ecdr/intermediate_monthly.py
@@ -204,7 +204,7 @@ CDR_SEAICE_CONC_QA_FLAG_DAILY_BITMASKS = OrderedDict(
 )
 
 # TODO: rename. This is actually a bit mask (except the fill_value)
-CDR_SEAICE_CONC_MONTHLY_QA_FLAG_BITMASKS = OrderedDict(
+CDR_SEAICE_CONC_MONTHLY_QA_FLAG_BITMASKS_NORTH = OrderedDict(
     {
         "average_concentration_exceeds_0.15": 1,
         "average_concentration_exceeds_0.30": 2,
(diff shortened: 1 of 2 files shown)


seaice_ecdr/intermediate_monthly.py
```diff
@@ -217,6 +217,18 @@ def get_daily_ds_for_month(
...
+        "average_concentration_exceeds_0.15": 1,
+        "average_concentration_exceeds_0.30": 2,
+        "at_least_half_the_days_have_sea_ice_conc_exceeds_0.15": 4,
+        "at_least_half_the_days_have_sea_ice_conc_exceeds_0.30": 8,
+        "invalid_ice_mask_applied": 16,
+        "at_least_one_day_during_month_has_spatial_interpolation": 32,
+        "at_least_one_day_during_month_has_temporal_interpolation": 64,
+    }
+)
+
```
trey-stafford, line 231:
```suggestion
CDR_SEAICE_CONC_MONTHLY_QA_FLAG_BITMASKS_SOUTH = CDR_SEAICE_CONC_MONTHLY_QA_FLAG_BITMASKS_NORTH.copy()
CDR_SEAICE_CONC_MONTHLY_QA_FLAG_BITMASKS_SOUTH.pop("at_least_one_day_during_month_has_melt_detected")

```

This makes it a little more clear that the SH bitmasks are the same as NH, just excluding the melt onset detection bitmask.

seaice_ecdr/intermediate_monthly.py
```diff
@@ -319,34 +326,30 @@ def calc_cdr_seaice_conc_monthly_qa_flag(
...
+            flag_value=CDR_SEAICE_CONC_QA_FLAG_DAILY_BITMASKS["start_of_melt_detected"],
+        ).any(dim="time")
+        cdr_seaice_conc_monthly_qa_flag = cdr_seaice_conc_monthly_qa_flag.where(
+            ~at_least_one_day_during_month_has_melt_detected,
+            cdr_seaice_conc_monthly_qa_flag
+            + qa_flag_bitmasks["at_least_one_day_during_month_has_melt_detected"],
+        )
+        monthly_qa_valid_range = (np.uint8(0), np.uint8(255))
+    else:
+        monthly_qa_valid_range = (np.uint8(0), np.uint8(127))
```
trey-stafford, line 344: I think we could calculate the range without needing to hard-code this here, letting the bitmask constants drive the result:

```
monthly_qa_max = sum([value for value in qa_flag_bitmasks.values()]
monthly_qa_valid_range = (np.uint8(0), np.uint8(monthly_qa_max))
```

trey-stafford (changes requested):
Looks good - thanks for getting this done! I have one suggestion I think we should apply, and one other idea to avoid hard-coding the max QA bitmask value. Otherwise looks good!

seaice_ecdr/intermediate_monthly.py
trey-stafford: 53accf2

PR #175, opened 2024-11-07
allow melt from not-last-day-of-month

sc0tts:


seaice_ecdr/intermediate_monthly.py
```diff
@@ -477,10 +477,20 @@ def calc_cdr_melt_onset_day_monthly(
...
+    max_date = daily_melt_onset_for_month.time.max()
+    if daily_melt_onset_for_month.time.dtype == np.datetime64(1, "ns").dtype:
+        unix_epoch = np.datetime64(0, "s")
+        one_second = np.timedelta64(1, "s")
+        for dt64 in daily_melt_onset_for_month.time.data:
+            seconds_since_epoch = (dt64 - unix_epoch) / one_second
+            date = dt.datetime.utcfromtimestamp(seconds_since_epoch).date()
+            if date.strftime("%j") == "244":
+                max_date = dt64
+                logger.info(f"Found non-max date with melt: {max_date}")
```
trey-stafford, line 491: I might be missing something here, but I think conditional & for-loop can be simplified to:

```
doy_str_list = [date.strftime("%j") for date in daily_melt_onset_for_month.time.dt.date.values]
try:
    # If day 244 exists in the index, we use that as the max date.
    day_244_idx = doy_str_list.index("244")
    max_date = daily_melt_onset_for_month.time[day_244_idx].values
except ValueError:
    # Day 244 does not exist in the index. Use the max date in the list.
    max_date = daily_melt_onset_for_month.time.max()

```

trey-stafford (changes requested):
one question and one possible suggestion I'd like your feedback on!

seaice_ecdr/intermediate_monthly.py
trey-stafford: Actually I think this is a little better:

```
doy_str_list = [date.strftime("%j") for date in daily_melt_onset_for_month.time.dt.date.values]
if "244" in doy_str_list:
    day_244_idx = doy_str_list.index("244")
    max_date = daily_melt_onset_for_month.time[day_244_idx].values
else:
    max_date = daily_melt_onset_for_month.time.max()
```

trey-stafford: Lets use the doy max set in `melt.py` instead of hard-coding it here.

trey-stafford (approved):

PR #177, opened 2024-11-13
Fixups for prod

trey-stafford:
Finalize production deployment: fix bugs as they come up during processing w/ ops.

We will merge this PR after we're happy with production outputs.


sc0tts (approved):
Minor tweaks; 
Good catch on the cwd kwarg for subprocess.run()

PR #182, opened 2025-07-03
Add support for producing monthly NRT from 0802

### LINKED ISSUE #181: Support producing monthly NRT CDR with 0802
Focus on producing monthly NRT CDR from 0802 data.

We should retain the ability to produce prototype data files, but we want to promote amsr2 out of prototype.

trey-stafford:
Resolves #181 

Retains the code to produce a prototype group, but makes it optional. This still feels a little hacky, and might be nice to just remove the code to support the "prototype" group altogether, but it may yet come in handy w/ e.g., AMSR3


seaice_ecdr/cli/monthly_nrt.py
trey-stafford, line 1: TODO: revisit option to spec nrt platform. We want to support AM3 as a prototype platform eventually.

PR #193, opened 2025-08-14
Ancillary thresholds variables

trey-stafford:


trey-stafford:
We decided not to apply the Seki method to the DMSP data, so this PR will need to be updated to reflect that.

PR that switches to applying Seki thresholds to AMSR2 data instead: https://github.com/nsidc/seaice_ecdr/pull/196

